PR #2, opened 2018-11-29
New API

peterbourgon:
Wildly different approach, same net effect.


parse.go
```diff
@@ -0,0 +1,190 @@
...
+			}
+
+			var key string
+			{
+				key = strings.ToUpper(f.Name)
+				key = envVarReplacer.Replace(key)
+				key = strings.ToUpper(c.envVarPrefix) + "_" + key
+			}
+			if value := os.Getenv(key); value != "" {
+				for _, individual := range strings.Split(value, ",") {
```
omidaladini: Just bumped into this; why do env vars get split by comma?

omidaladini: I see. Yes I did have an env var which was a comma separated list of things :) I was under the assumption that string flags and their respective env vars behave similarly. Following up on the PR: https://github.com/peterbourgon/ff/pull/39

PR #5, opened 2018-12-06
Group parser docs

carlmjohnson:
The docs are easier to read if PlainParser is under ConfigFileParser.


peterbourgon:
Hmm, this is half a step too clever for my taste.

PR #11, opened 2018-12-10
Add JSONParser

### LINKED ISSUE #6: JSON config parser?
It would be pretty trivial to write. If I open a PR, would you accept it, or would you prefer people write their own config parsers?

carlmjohnson:
Fixes #6.

### DIFF
--- a/json.go
+++ b/json.go
@@ -62,6 +62,7 @@ func stringifyValue(val interface{}) (string, error) {
 		return v.String(), nil
 	case bool:
 		return strconv.FormatBool(v), nil
+	default:
+		return "", errors.Errorf("could not convert %q (type %T) to string", val, val)
 	}
-	return "", errors.Errorf("could not convert %q (type %T) to string", val, val)
 }
--- a/json.go
+++ b/json.go
@@ -1,0 +1,67 @@
+package ff
+
+import (
+	"encoding/json"
+	"io"
+	"strconv"
+
+	"github.com/pkg/errors"
+)
+
+// JSONParser is a parser for config files in JSON format. Input should be
+// an object. The object's keys are treated as flag names, and the object's
+// values as flag values. If the value is an array, the flag will be set
+// multiple times.
+func JSONParser(r io.Reader, set func(name, value string) error) error {
+	var m map[string]interface{}
+	d := json.NewDecoder(r)
+	// Must set UseNumber for stringifyValue to work
+	d.UseNumber()
+	err := d.Decode(&m)
+	if err != nil {
+		return errors.Wrap(err, "error parsing JSON config")
+	}
+	for key, val := range m {
+		values, err := stringifySlice(val)
+		if err != nil {
+			return errors.Wrap(err, "error parsing JSON config")
+		}
+		for _, value := range values {
+			if err := set(key, value); err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+func stringifySlice(val interface{}) ([]string, error) {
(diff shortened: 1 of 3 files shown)


parse.go
```diff
@@ -41,7 +41,7 @@ func Parse(fs *flag.FlagSet, args []string, options ...Option) error {
 		}
 		defer f.Close()
 
-		c.configFileParser(f, func(name, value string) error {
+		err = c.configFileParser(f, func(name, value string) error {
```
peterbourgon: Ah, thanks.

json_test.go
```diff
@@ -0,0 +1,113 @@
...
+			name: "Key with array of values",
+			args: []string{},
+			file: `
+                {
+                    "s": ["t", "s"],
+                    "i": ["11", "10"],
+                    "b": [false, true],
+                    "d": ["10m", "5s"]
+                }
+            `,
```
peterbourgon: Nit: can you please fix this up with tabs for indention?

json_test.go
```diff
@@ -0,0 +1,113 @@
...
+			want: want{"s", 10, true, 5 * time.Second, ""},
+		},
+		{
+			name: "bad JSON file",
+			args: []string{},
+			file: `{`,
+			want: want{d: time.Second, err: "error parsing JSON config"},
```
peterbourgon: Is there a way to improve this, to leverage the ParseError type, so you're not comparing strings?

json.go
```diff
@@ -0,0 +1,67 @@
...
+// multiple times.
+func JSONParser(r io.Reader, set func(name, value string) error) error {
+	var m map[string]interface{}
+	d := json.NewDecoder(r)
+	// Must set UseNumber for stringifyValue to work
+	d.UseNumber()
+	err := d.Decode(&m)
+	if err != nil {
+		return errors.Wrap(err, "error parsing JSON config")
+	}
```
peterbourgon: Nit: just to make this a bit tighter...

```go
d := json.NewDecoder(r)
d.UseNumber() // stringify expects json.Numbers
var m map[string]interface{}
if err := d.Decode(&m); err != nil {
    return ParseError{Err: err}
}
```

json.go
```diff
@@ -0,0 +1,67 @@
...
+func stringifyValue(val interface{}) (string, error) {
+	switch v := val.(type) {
+	case string:
+		return v, nil
+	case json.Number:
+		return v.String(), nil
+	case bool:
+		return strconv.FormatBool(v), nil
+	}
+	return "", errors.Errorf("could not convert %q (type %T) to string", val, val)
```
peterbourgon: Extremely minor nit: can you move this to a `default` clause?

json.go
```diff
@@ -0,0 +1,67 @@
+package ff
```
peterbourgon: Honest question: do you think it's better to move this to its own ffjson subpackage? Pro: symmetry with potential future ffyaml subpackage. Con: this one only uses stdlib, so not strictly necessary?

carlmjohnson: Might make more sense.

json_test.go
carlmjohnson: Which ParseError? You mean the errors in encoding/json? For a test, I'm inclined to be happy with just a quick and dirty string test.

json.go
carlmjohnson: If I'm creating a public facing error type, then yeah, this should probably be in a new package.

peterbourgon:
May end up breaking this out but fine for now I think.

PR #12, opened 2018-12-17
Allow for environment variables with no prefix.

jmhobbs:
I wanted to use ff to replace an internal tool which uses un-prefixed environment variables.

I'm not sure if this is a behavior you want to support, but I created a patch to discuss it.

Thanks!

### DIFF
--- a/parse.go
+++ b/parse.go
@@ -65,7 +65,7 @@ func Parse(fs *flag.FlagSet, args []string, options ...Option) error {
 		provided[f.Name] = true
 	})
 
-	if c.envVarPrefix != "" {
+	if c.envVarPrefix != "" || c.envVarNoPrefix {
 		var errs []string
 		fs.VisitAll(func(f *flag.Flag) {
 			if provided[f.Name] {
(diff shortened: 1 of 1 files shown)


parse.go
```diff
@@ -141,6 +144,18 @@ func WithEnvVarPrefix(prefix string) Option {
 	}
 }
 
+// WithEnvVarNoPrefix tells parse to look in the environment for variables with
+// no prefix.  Flag names are converted to environment variables by
+// capitalizing them, and replacing separator characters like periods or hyphens
+// with underscores. Additionally, if the environment variable's value contains
+// commas, each comma-delimited token is treated as a separate instance of the
+// associated flag name.
```
peterbourgon: Rather than duplicating the explanation text, I think it's OK to say something like "See WithEnvVarPrefix for an explanation of how flag names are converted to environment variables names."

jmhobbs: Can do.

jmhobbs: Replaced.

peterbourgon:
Thanks!

PR #13, opened 2018-12-18
Fix handling env vars with comma-separated values

jforns:
In the handling of environment variables values are split by comma and `FlagSet::Set` is called for each part, when the expected behavior is to call `Set` with the untouched value.

This PR removes the split by comma and adds tests for comma-separated values in env vars, configuration files and flags.

### DIFF
--- a/parse.go
+++ b/parse.go
@@ -81,10 +81,8 @@ func Parse(fs *flag.FlagSet, args []string, options ...Option) error {
 				}
 			}
 			if value := os.Getenv(key); value != "" {
-				for _, individual := range strings.Split(value, ",") {
-					if err := fs.Set(f.Name, strings.TrimSpace(individual)); err != nil {
-						errs = append(errs, errors.Wrapf(err, "error setting flag %q from env var %q", f.Name, key).Error())
-					}
+				if err := fs.Set(f.Name, strings.TrimSpace(value)); err != nil {
+					errs = append(errs, errors.Wrapf(err, "error setting flag %q from env var %q", f.Name, key).Error())
 				}
 			}
 		})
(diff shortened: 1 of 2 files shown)


peterbourgon:
I'm sorry, I don't understand this PR. What bug do you think this is fixing? I _only_ want to split on commas in env vars, not in explicit flags or config file values.

jforns:
I will explain my use case, as it might defer from what you intended.

I integrated the package in a service that receives a few configuration options, one of which is a `map[string]string`.
At that point the service could only be configured with flags, and the one with the value of the map was encoded in a string with format `key1:value1,key2:value2` and so on. 
A type that implemented `flag:Value` received the configuration option, it split the string by commas, then by colons and finally filled the map with the pairs.
Then when I integrated the `ff` package I expected it to have a behavior that was consistent with the `flag` package, but the split in https://github.com/peterbourgon/ff/blob/master/parse.go#L84 prevented it to be like so.

IMHO the responsibility of parsing the content of a configuration option value should be on the `flag:Value` instead of on `ff:Parse` and the outcome should be the same regardless of how a configuration option is passed.

Please let me know your thoughts.

peterbourgon:
> IMHO the responsibility of parsing the content of a configuration option value should be on the flag:Value instead of on ff:Parse and the outcome should be the same regardless of how a configuration option is passed.

I want the ability to specify the same flag with different values multiple times. This is possible with explicit flags by repeating the flag; it's possible with most/all config file formats by using an e.g. array of values; it's not possible with environment vars except by parsing the string.

I'm happy to entertain workarounds that don't change this behavior, and don't complect the code or API too much.

PR #17, opened 2019-01-08
PlainParser: hash in value isn't a comment

### LINKED ISSUE #16: PlainParser: # in config file value is interpreted as comment and truncates
If you have a config file value that includes `#`, the value is truncated because PlainParser interprets the rest of the line as a comment, which is undesirable.

For example:
```
password abc#123
```

`password` is interpreted as `abc`, not `abc#123`.

I think there's probably no legitimate cases where a comment would be added without a space beforehand, so if the `strings.IndexRune` call on [line 188](https://github.com/peterbourgon/ff/blob/979ab8ea50a3d6af940f159d81037b01ebe5bac1/parse.go#L188) instead were `string.Index(value, " #")`, that'd solve this issue and I think still cover all legit uses.

peterbourgon:
Closes #16.

### DIFF
--- a/parse.go
+++ b/parse.go
@@ -185,7 +185,7 @@ func PlainParser(r io.Reader, set func(name, value string) error) error {
 			name, value = line[:index], strings.TrimSpace(line[index:])
 		}
 
-		if i := strings.IndexRune(value, '#'); i >= 0 {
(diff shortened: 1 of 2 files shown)


joeshaw:
LGTM, thanks! 👍

PR #21, opened 2019-02-13
Remove pkg/errors dependency

### LINKED ISSUE #20: Remove github.com/pkg/errors dependency
It would be cool if this package could be totally dependency-free, by removing the dependency on pkg/errors. We only use the `Wrap` functionality, and not to any sophisticated end; it could easily be re-implemented, or elided altogether in favor of plain string-wrapping, e.g.

```diff
-    return errors.Wrap(err, "error doing foo")
+    return fmt.Errorf("error doing foo: %v", err)
```

peterbourgon:
Closes #20.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,1 @@
 module github.com/peterbourgon/ff
-
-require (
-	github.com/BurntSushi/toml v0.3.1
-	github.com/pkg/errors v0.8.1
-)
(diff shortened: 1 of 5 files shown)


peterbourgon:
My mod-fu is weak. @flowchartsman, would you mind putting your eyes on this and verifying it makes sense? /cc also @crawshaw.

flowchartsman:
I'm no authority on the matter, but LGTM, especially since the TOML lib version and hash are identical.

I usually just use the `%s` verb for errors since `fmt.pp.handleMethods` treats them the same at the error interface, but that's just a stylistic thing.

Issue #19, opened 2019-02-13
toml dependency and modules


crawshaw:
This is a very nice package that I'm considering using.

I don't want to introduce a toml dependency in my project. This is easily fixed as I will vendor this code and remove the toml-related code when I do. But it leaves me wondering: is there a better way to do this under modules? I thought about sending a PR with the toml code moved to a fftoml package, but that still leaves the module dependency. Is there some other way to do this?

flowchartsman:
Can't you just move all toml code to a subdirectory and initialize that as its own module?

```
$ git clone --depth=1 --branch=master git@github.com:peterbourgon/ff
$ rm -rf .git go.mod go.sum
$ mkdir toml
$ mv toml*.go toml
$ cd toml
$ # EDIT EDIT
$ head -n1 toml.go
package toml // import "github.com/peterbourgon/ff/toml"
$ go mod init
go: creating new go.mod: module github.com/peterbourgon/ff/toml
$ go build ./...
go: finding github.com/pkg/errors v0.8.1
go: finding github.com/BurntSushi/toml v0.3.1
go: downloading github.com/pkg/errors v0.8.1
go: downloading github.com/BurntSushi/toml v0.3.1
$ cat go.mod
module github.com/peterbourgon/ff/toml

require (
	github.com/BurntSushi/toml v0.3.1
	github.com/pkg/errors v0.8.1
)
$ cat go.sum
github.com/BurntSushi/toml v0.3.1 h1:WXkYYl6Yr3qBf1K79EBnL4mak0OimBfB0XUf9Vl28OQ=
github.com/BurntSushi/toml v0.3.1/go.mod h1:xHWCNGjB5oqiDr8zfno3MHue2Ht5sIBksp03qcyfWMU=
github.com/pkg/errors v0.8.1 h1:iURUrRGxPUNPdy5/HRSm+Yj6okJ6UtLINN0Q9M4+h3I=
github.com/pkg/errors v0.8.1/go.mod h1:bwawxfHBFNV+L2hUp1rHADufV3IMtnDRdf1r5NINEl0=
$ cd ..
$ vi parse.go
$ # EDIT EDIT (not considering testing changes)
$ head -n1 parse.go
package ff // import "github.com/peterbourgon/ff"
$ go mod init
go: creating new go.mod: module github.com/peterbourgon/ff
$ go build ./...
$ cat go.mod
module github.com/peterbourgon/ff

require github.com/pkg/errors v0.8.1
$ cat go.sum
[...]

PR #22, opened 2019-02-14
Move TOML parser to fftoml package.

crawshaw:


crawshaw:
cc @flowchartsman

flowchartsman:
Import comment should be unnecessary in this case. I only used it to make it work in my non-gopath.  I believe it will be fine without it.

flowchartsman:
And, again, I'm not sure how the dependency on ff will work if the rest of the lib gets updated independently, since test dependencies are not separate from build dependencies. That's something to look into. Not a lot of sub-modules about right now.

flowchartsman:
I mean, you could also argue that they should be tightly-coupled...

crawshaw:
I couldn't generate the modules files without the import path.

I agree there's something unusual about having fftoml depend on the ff module, but I think it makes sense. And as long as anyone submitting a PR runs go test ./... they'll end up tightly coupling the submodule.

(I'm very new to modules, I do most of my development under GOPATH still to avoid confusing my editor setup. Following a late adopter strategy on this one.)

peterbourgon:
Respecting my understanding of Filo's suggestion in #19, I propose this patch:

```diff
diff --git a/fftoml/fftoml.go b/fftoml/fftoml.go
index e5685b5..b52b96d 100644
--- a/fftoml/fftoml.go
+++ b/fftoml/fftoml.go
@@ -1,4 +1,4 @@
-package fftoml // import "github.com/peterbourgon/ff/fftoml"
+package fftoml
 
 import (
 	"fmt"
diff --git a/fftoml/fftoml_test.go b/fftoml/fftoml_test.go
index 47cfcd4..81e513f 100644
--- a/fftoml/fftoml_test.go
+++ b/fftoml/fftoml_test.go
@@ -1,4 +1,4 @@
-package fftoml
+package fftoml_test
 
 import (
 	"flag"
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	"github.com/peterbourgon/ff"
+	"github.com/peterbourgon/ff/fftoml"
 )
 
 func TestParser(t *testing.T) {
@@ -75,7 +76,7 @@ func TestParser(t *testing.T) {
 				f.Write([]byte(testcase.file))
 				f.Close()
 
-				options = append(options, ff.WithConfigFile(f.Name()), ff.WithConfigFileParser(Parser))
+				options = append(options, ff.WithConfigFile(f.Name()), ff.WithConfigFileParser(fftoml.Parser))
 			}
 
 			err := ff.Parse(fs, testcase.args, options...)
diff --git a/fftoml/go.mod b/fftoml/go.mod
deleted file mode 100644
index 6e26c6a..0000000
--- a/fftoml/go.mod
+++ /dev/null
@@ -1,6 +0,0 @@
-module github.com/peterbourgon/ff/fftoml
-
-require (
-	github.com/BurntSushi/toml v0.3.1
-	github.com/peterbourgon/ff v1.1.0
-)
diff --git a/go.mod b/go.mod
index 4aa1520..5cb823f 100644
--- a/go.mod
+++ b/go.mod
@@ -1 +1,3 @@
 module github.com/peterbourgon/ff
+
+require github.com/BurntSushi/toml v0.3.1
diff --git a/fftoml/go.sum b/go.sum
similarity index 50%
rename from fftoml/go.sum
rename to go.sum
index f2d5441..9cb2df8 100644
--- a/fftoml/go.sum
+++ b/go.sum
@@ -1,4 +1,2 @@
 github.com/BurntSushi/toml v0.3.1 h1:WXkYYl6Yr3qBf1K79EBnL4mak0OimBfB0XUf9Vl28OQ=
 github.com/BurntSushi/toml v0.3.1/go.mod h1:xHWCNGjB5oqiDr8zfno3MHue2Ht5sIBksp03qcyfWMU=
-github.com/peterbourgon/ff v1.1.0 h1:mz0/c1gtUwz/MsyJs3GwPjFSTFlD4UEwKEPw5Z+0cm8=
-github.com/peterbourgon/ff v1.1.0/go.mod h1:P2CDTJbjip+iAWS2jnkp6uKBX2Bcvc/K4p5j47CTRv0=
diff --git a/parse.go b/parse.go
index 89f4591..69e5c67 100644
--- a/parse.go
+++ b/parse.go
@@ -1,4 +1,4 @@
-package ff // import "github.com/peterbourgon/ff"
+package ff
 
 import (
 	"bufio"
```

crawshaw:
@peterbourgon done.

I have a slight naive preference for the sub-module approach, but I'm new to modules in anger and so am happy to go along for the ride.

FiloSottile:
As this is a breaking change, and as v1 has already been tagged, the module should be bumped to v2. (Regardless of whether a submodule is made.)

crawshaw:
oh no. can we take a mulligan? go back to v0?

FiloSottile:
@peterbourgon If you want to change anything else, the time is now!

crawshaw:
If there are other changes to accumulate, the version bump can be in a later PR, no? If I understand correctly, Go modules will use the last tagged version until a new one is tagged.

fftoml/fftoml_test.go
```diff
@@ -9,9 +9,12 @@ import (
 	"strings"
 	"testing"
 	"time"
+
+	"github.com/peterbourgon/ff/v2"
+	"github.com/peterbourgon/ff/v2/fftoml"
```
peterbourgon: I'm not happy with changing the import path. Is there no way around this?

peterbourgon: This is sufficiently unacceptable to me as an outcome that I am inclined to break compatibility promises and stay on v0/v1 in perpetuity. Package identifier and version are and ought to be independent variables, modules' insistence on co-mingling them is an unacceptable design choice.

crawshaw:
Hey, good news! The last release was v1.1.0, which was done before the commit to add the TOML parser. So this isn't a breaking change as far as modules is concerned, and we can stick to v1.

peterbourgon:
```go
defer angst() // :)
```

crawshaw:
This is definitely a squash and merge. 😅

fftoml/fftoml_test.go
FiloSottile: (While I personally dislike the versioned import path for a number of reasons, I would see breaking SemVer as a signal that I can't rely on a module to be easy to keep up to date, and I would be discouraged from using it. A standard way to send that signal is to stay on v0.)

denisbrodbeck:
Any plans with a new tag for the recent changes?

Tried to import `github.com/peterbourgon/ff` and `github.com/peterbourgon/ff/fftoml` into a new project following the [go docs](https://godoc.org/github.com/peterbourgon/ff), ended scratching my had wondering why go won't find `github.com/peterbourgon/ff/fftoml`.
Took some time until I noticed that `go mod` took the latest tag `v1.1.0` :disappointed: Pinning `github.com/peterbourgon/ff` to the last commit on master now instead.

peterbourgon:
Sorry for that, oversight on my part, I've just tagged v1.2.0.

denisbrodbeck:
@peterbourgon Thank you, that was fast :+1:

PR #26, opened 2019-06-23
Add ffyaml

### LINKED ISSUE #9: Yaml config parser
At our company we heavily use yaml for configurations between env. 

I would like to try out using this package on one of our new services. Would you be open to accepting a yaml config parser implementation ?

yjp20:
closes #9

### DIFF
--- a/ffyaml/yaml.go
+++ b/ffyaml/yaml.go
@@ -1,0 +1,22 @@
+package ffyaml
+
+import (
+	"fmt"
+	"io"
+
+	"gopkg.in/yaml.v2"
+)
+
+func Parser(r io.Reader, set func(name, value string) error) error {
+	var m map[string]string
+	d := yaml.NewDecoder(r)
+	if err := d.Decode(&m); err != nil && err != io.EOF {
+		return fmt.Errorf("error parsing YAML config: %v", err)
+	}
+	for key, val := range m {
+		if err := set(key, val); err != nil {
+			return err
+		}
+	}
+	return nil
+}
--- a/ffyaml/yaml_test.go
+++ b/ffyaml/yaml_test.go
@@ -1,0 +1,125 @@
+package ffyaml_test
+
+import (
+	"flag"
+	"fmt"
+	"math/rand"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/peterbourgon/ff"
+	"github.com/peterbourgon/ff/ffyaml"
+)
+
+func TestYAMLParser(t *testing.T) {
+	type want struct {
+		s   string
+		i   int
+		b   bool
+		d   time.Duration
+		f   float64
+		err string
+	}
+
+	for _, testcase := range []struct {
+		name string
+		args []string
+		file string
+		want want
+	}{
+		{
+			name: "empty input",
+			args: []string{},
+			file: ``,
+			want: want{d: time.Second},
+		},
+		{
+			name: "basic KV pairs",
+			args: []string{},
+			file: `
+s: hello
+i: 10
+b: true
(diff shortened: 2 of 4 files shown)


peterbourgon:
@yjp20 Thank you! One failure:

```
ffyaml/yaml.go:6:2: imported and not used: "strconv"
```

peterbourgon:
@crawshaw As a previously casually interested party to this issue, do you have any opinions here?

yjp20:
I don't know what happened, but I thought I made sure to test before pushing. I've fixed a few mistakes and it seems to be ok now.

ffyaml/yaml.go
```diff
@@ -0,0 +1,22 @@
...
+	"fmt"
+	"io"
+
+	"gopkg.in/yaml.v2"
+)
+
+func Parser(r io.Reader, set func(name, value string) error) error {
+	var m map[string]string
+	d := yaml.NewDecoder(r)
+	if err := d.Decode(&m); err != nil && err.Error() != "EOF" {
```
peterbourgon: You should be able to check for io.EOF directly here, e.g.

```go
if err := d.Decode(&m); err != nil && err != io.EOF {
```

ffyaml/yaml.go
```diff
@@ -0,0 +1,22 @@
...
+		return fmt.Errorf("error parsing YAML config: %v", err)
+	}
+	for key, val := range m {
+		if err := set(key, val); err != nil {
+			return err
+		}
+	}
```
peterbourgon: How will you support flags that can be set multiple times? Can you add support for something like YAML arrays or lists, similar to [this bit of code in the JSON parser](https://github.com/peterbourgon/ff/blob/52d851567994fcca73bf3668fc3099cfa7be3768/json.go#L22)?

ffyaml/yaml.go
peterbourgon: Also, do you know how to trigger [the branch in Decode that returns io.EOF](https://github.com/go-yaml/yaml/blob/51d6538a90f86fe93ac480b35f37b2be17fef232/yaml.go#L122-L125)? It would be good to add a test case for that.

ffyaml/yaml.go
yjp20: I know it's possible... but I don't really see the point. Doesn't fs only support singular values?

peterbourgon: Flags can be repeated, and if the value supports it, each instance of the flag can be saved separately. None of the default flag value types support that but it’s easy to write your own which does. I use this sort of thing frequently.

ffyaml/yaml_test.go
```diff
@@ -0,0 +1,125 @@
...
+			},
+		},
+		{
+			name: "bad YAML file",
+			args: []string{},
+			file: `	s:`,
+			want: want{d: time.Second, err: "error parsing YAML config"},
+		},
+		{
+			name: "EOF example",
```
cbandy: This looks the same as the test case on L34.

crawshaw:
I missed this when I was added (bad email config). The package separation looks good.

peterbourgon:
Took it to the finish line in #31, thanks for your contribution!

ffyaml/yaml.go
carlmjohnson: See also https://godoc.org/github.com/carlmjohnson/flagext#Strings

yjp20:
Sorry about not finishing this myself, I was planning on doing it but other stuff got a hold of me for a while. I would like to add that I think you were definitely "right" about this way of handling configuration, I've been using it recently and it just works.

peterbourgon:
@yjp20 No worries and great to hear :)

PR #39, opened 2019-08-28
Test case to clarify env var behavior

peterbourgon:
@omidaladini This clarifies my intent re: your earlier question. Would you like a different behavior?


omidaladini:
Thanks Peter; Just realized what your intention was. Great to have it documented on a test case. 

Was wondering, since splitting is specific to certain kind of flags, such as StringSlice, shouldn't it be `StringSlice.Set`'s job to split the flag accordingly? Then we'd have:


```
		{
			name: "repeated env vars",
			env:  map[string]string{"TEST_PARSE_S": "one,two,three", "TEST_PARSE_X": "one,two,three"},
			want: fftest.Vars{S: "one,two,three", X: []string{"one", "two", "three"}},
		},
```

peterbourgon:
I approached it from a different angle, which is that both commandline flags and config files have a native way to give a flag an "array" of values, and that env vars don't, so the best approach would be to define a global way to get that behavior out of env vars, rather than imbuing particular values with special parse logic.

Hence the last sentence of [the WithEnvVarPrefix doc comment](https://github.com/peterbourgon/ff/blob/f3eac38b49b92570fa4dd9c8b03b1a5d61e9bc90/parse.go#L134-L139):

> Additionally, if the environment variable's value contains commas, each comma-delimited token is treated as a separate instance of the associated flag name.

But, your way of looking at things is also totally valid.

I'm inclined to add this patch:

```diff
diff --git a/parse.go b/parse.go
index 0d8c166..e090e20 100644
--- a/parse.go
+++ b/parse.go
@@ -80,8 +80,14 @@ func Parse(fs *flag.FlagSet, args []string, options ...Option) error {
 				}
 			}
 			if value := os.Getenv(key); value != "" {
-				for _, individual := range strings.Split(value, ",") {
-					if err := fs.Set(f.Name, strings.TrimSpace(individual)); err != nil {
+				var values []string
+				if c.envVarIgnoreCommas {
+					values = []string{value}
+				} else {
+					values = strings.Split(value, ",")
+				}
+				for _, v := range values {
+					if err := fs.Set(f.Name, v); err != nil {
 						errs = append(errs, fmt.Sprintf("error setting flag %q from env var %q: %v", f.Name, key, err))
 					}
 				}
@@ -102,6 +108,7 @@ type Context struct {
 	configFileParser   ConfigFileParser
 	envVarPrefix       string
 	envVarNoPrefix     bool
+	envVarIgnoreCommas bool
 }
 
 // Option controls some aspect of parse behavior.
@@ -134,9 +141,7 @@ func WithConfigFileParser(p ConfigFileParser) Option {
 // WithEnvVarPrefix tells parse to look in the environment for variables with
 // the given prefix. Flag names are converted to environment variables by
 // capitalizing them, and replacing separator characters like periods or hyphens
-// with underscores. Additionally, if the environment variable's value contains
-// commas, each comma-delimited token is treated as a separate instance of the
-// associated flag name.
+// with underscores.
 func WithEnvVarPrefix(prefix string) Option {
 	return func(c *Context) {
 		c.envVarPrefix = prefix
@@ -152,6 +157,17 @@ func WithEnvVarNoPrefix() Option {
 	}
 }
 
+// WithEnvVarIgnoreCommas tells parse to ignore commas in environment variable
+// values, treating the complete value as a single string passed to the
+// associated flag. By default, if an environment variable's value contains
+// commas, each comma-delimited token is treated as a separate instance of the
+// associated flag.
+func WithEnvVarIgnoreCommas(ignore bool) Option {
+	return func(c *Context) {
+		c.envVarIgnoreCommas = ignore
+	}
+}
+
 // ConfigFileParser interprets the config file represented by the reader
 // and calls the set function for each parsed flag pair.
 type ConfigFileParser func(r io.Reader, set func(name, value string) error) error
```

WDYT?

omidaladini:
> I approached it from a different angle, which is that both commandline flags and config files have a native way to give a flag an "array" of values, and that env vars don't, so the best approach would be to define a global way to get that behavior out of env vars, rather than imbuing particular values with special parse logic.

Makes a lot of sense.

The patch looks great and would help with my case!

Thanks a lot 👍

PR #40, opened 2019-08-29
support nested flags parse/setup

gobwas:
Hi! First of all, thank you for such good interpretation of how Go programs should be configured.

This PR brings an ability to export flags by some libraries/components of the program which support some conventions of exporting their parameters:

```go
package main

import (
    "github.com/peterbourgon/ff"

    "my/cool/lib/libconfig"
)

func main() {
    fs := flag.NewFlagSet("main", 0)
    ...
    ff.Subset(fs, "lib", func(fs *flag.FlagSet) {
        _ = libconfig.Export(fs)
    })
    ff.Parse(fs, ...)
}
```

```go
package libconfig

import "my/cool/lib"

type Config interface {
    StringVar(p *string, name, value, usage string)
}

func Export(c Config) (ret *lib.Config) {
    ret = new(lib.Config)
    c.StringVar(&ret.Foo, "foo", "bar", "usage")
    return
}
```

This will allow to run our program as `binary -lib.foo xxx` and so on.

### DIFF
--- a/fftest/vars.go
+++ b/fftest/vars.go
@@ -25,6 +25,8 @@ func Pair() (*flag.FlagSet, *Vars) {
 	fs.DurationVar(&v.D, "d", 0*time.Second, "time.Duration")
 	fs.Var(&v.X, "x", "collection of strings (repeatable)")
 
+	fs.StringVar(&v.SubS, "sub.s", "", "string")
+
 	return fs, &v
 }
 
@@ -36,6 +38,8 @@ type Vars struct {
 	B bool
 	D time.Duration
 	X StringSlice
+
+	SubS string
 
 	// ParseError should be assigned as the result of Parse in tests.
 	ParseError error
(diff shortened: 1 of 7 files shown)


peterbourgon:
Thanks for the interest. Unfortunately, flag "subsets" are not a pattern I would like ff to support.

If this is your use case, you can write an adapter which passes an intermediate flag.FlagSet to libconfig.Export. Then, you can introspect each flag for its name, and register a corresponding e.g. prefixed flag in your own flag.FlagSet.

gobwas:
Hi,

That is exactly what does [ff.Subset() from the PR](https://github.com/peterbourgon/ff/pull/40/files#diff-090598d65549267a01d4ed4132809255R105) :)

Any way, Subset() function is not the main point of this request for changes – the main idea is to allow parsing nested fields wihtin yaml or toml parsers. Currently it is impossible.

peterbourgon:
> The main idea is to allow parsing nested fields wihtin yaml or toml parsers. Currently it is impossible.

Yes, this is by design.

anarcher:
Hi, I just wonder why the ff doesn't' allow nested fields by design? :->

PR #44, opened 2019-09-19
Add IgnoreUndefined option

yjp20:
I think it might be useful for some people (like me) to have an option to ignore undefined keys. The best use case for this imo is when sharing a config over multiple binaries.

### DIFF
--- a/parse.go
+++ b/parse.go
@@ -41,7 +41,7 @@ func Parse(fs *flag.FlagSet, args []string, options ...Option) error {
 		defer f.Close()
 
 		err = c.configFileParser(f, func(name, value string) error {
-			if fs.Lookup(name) == nil {
+			if fs.Lookup(name) == nil && !c.ignoreUndefined {
 				return fmt.Errorf("config file flag %q not defined in flag set", name)
 			}
 
@@ -109,6 +109,7 @@ type Context struct {
 	envVarPrefix       string
 	envVarNoPrefix     bool
 	envVarIgnoreCommas bool
+	ignoreUndefined    bool
 }
 
 // Option controls some aspect of parse behavior.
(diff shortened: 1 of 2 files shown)


peterbourgon:
Just to be totally explicit: if a config file contains a line with a key that isn't part of the flagset, it should ignore that line totally?

parse.go
```diff
@@ -168,6 +169,14 @@ func WithEnvVarIgnoreCommas(ignore bool) Option {
 	}
 }
 
+// WithIgnoreUndefined tells parse to ignore keys that fail lookup during parse,
```
peterbourgon:
```suggestion
// WithIgnoreUndefined tells parse to ignore undefined flags that it encounters,
```

parse.go
```diff
@@ -168,6 +169,14 @@ func WithEnvVarIgnoreCommas(ignore bool) Option {
...
+// which would normally throw an error.
+func WithIgnoreUndefined() Option {
```
peterbourgon:
```suggestion
func WithIgnoreUndefined(ignore bool) Option {
```

parse.go
```diff
@@ -168,6 +169,14 @@ func WithEnvVarIgnoreCommas(ignore bool) Option {
...
+	return func(c *Context) {
+		c.ignoreUndefined = true
```
peterbourgon:
```suggestion
		c.ignoreUndefined = ignore
```

I know it's stupid, but it's consistent with the other bool options.

yjp20:
Fixed according to your comments :)

peterbourgon:
```
./parse_test.go:102:44: not enough arguments in call to ff.WithIgnoreUndefined
	have ()
	want (bool)
```

yjp20:
That's embarrassing... Didn't do my due diligence testing because I didn't have go 1.13. Lesson learned  I suppose.

PR #45, opened 2019-10-15
Fix WithIgnoreUndefined for config files

yjp20:
Found this option didnt work for configuration files.

### DIFF
--- a/parse_test.go
+++ b/parse_test.go
@@ -97,10 +97,22 @@ func TestParseBasics(t *testing.T) {
 			want: fftest.Vars{S: "one,two,three", X: []string{"one,two,three"}},
 		},
 		{
-			name: "WithIgnoreUndefined",
+			name: "WithIgnoreUndefined env",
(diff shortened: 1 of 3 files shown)


peterbourgon:
Can you add a test which fails under the previous implementation and succeeds in the fixed implementation?

yjp20:
Added the test cases

parse_test.go
```diff
@@ -97,11 +97,23 @@ func TestParseBasics(t *testing.T) {
...
+			name: "WithIgnoreUndefined file true",
+			file: "testdata/undefined.conf",
+			opts: []ff.Option{ff.WithIgnoreUndefined(true)},
+			want: fftest.Vars{S: "one"},
+		},
+		{
+			name: "WithIgnoreUndefined file false",
+			file: "testdata/undefined.conf",
+			opts: []ff.Option{ff.WithIgnoreUndefined(false)},
+			want: fftest.Vars{S: "", WantParseErrorString: "config file flag"},
```
peterbourgon:
```suggestion
			want: fftest.Vars{WantParseErrorString: "config file flag"},
```

PR #47, opened 2019-11-20
Closes #46

### LINKED ISSUE #46: Should `Parse` fail on missing config file?
Currently if `ff` is configured for using config file:
```
ff.Parse(fs, os.Args[1:], 
ff.WithConfigFile(appName+".conf"),
ff.WithConfigFileParser(ff.PlainParser))
```
And if that config file is missing, `Parse` fails. However it might be perfectly valid state, using defaults for the flags, and failing only if there is no default value for given flag.

Dragomir-Ivanov:


Dragomir-Ivanov:
This test breaks in non deterministic way.
```
 ~/ff    go test -race ./... -count 10                                                                                                                                       22.11.2019 (пт) 14:58:12 PST
ok  	github.com/peterbourgon/ff	1.124s
ok  	github.com/peterbourgon/ff/ffcli	1.067s
?   	github.com/peterbourgon/ff/ffcli/examples/objectctl	[no test files]
?   	github.com/peterbourgon/ff/fftest	[no test files]
ok  	github.com/peterbourgon/ff/fftoml	1.018s
--- FAIL: TestParser (0.00s)
    --- FAIL: TestParser/no_value (0.00s)
        ffyaml_test.go:67: I: want 123, have 0
FAIL
FAIL	github.com/peterbourgon/ff/ffyaml	0.034s
FAIL
 !  ~/ff    go test -race ./... -count 1                                                                                                                           1772ms  22.11.2019 (пт) 14:58:17 PST
ok  	github.com/peterbourgon/ff	1.018s
ok  	github.com/peterbourgon/ff/ffcli	1.009s
?   	github.com/peterbourgon/ff/ffcli/examples/objectctl	[no test files]
?   	github.com/peterbourgon/ff/fftest	[no test files]
ok  	github.com/peterbourgon/ff/fftoml	1.007s
ok  	github.com/peterbourgon/ff/ffyaml	1.011s

File: fftest/vars.go [hunk 1]
```diff
@@ -54,28 +54,24 @@ type Vars struct {
...
-		if want.WantParseErrorIs != nil && have.ParseError == nil {
-			return fmt.Errorf("want error (%v), have none", want.WantParseErrorIs)
-		}
-
-		if want.WantParseErrorString != "" && have.ParseError == nil {
-			return fmt.Errorf("want error (%q), have none", want.WantParseErrorString)
-		}
+	if want.WantParseErrorIs != nil && have.ParseError == nil {
+		return fmt.Errorf("want error (%v), have none", want.WantParseErrorIs)
+	}
```
peterbourgon: You can't change the Compare test helper in this way. If Parse returns an error, the result of the Parse against the flagset is undefined behavior, so you can't continue to test the other conditions. 

This is what's triggering the bug in ffyaml: it decodes the file into a map, and then ranges over the map to process each value, but ranging over a map in Go yields keys in random order, so sometimes it gets the `i` value first, sometimes the `s`.

Dragomir-Ivanov: Okay, let me try harder.

Dragomir-Ivanov: I don't know what I was thinking. Just reverting my change to that file, and everything works as expected. Thanks!

parse.go
```diff
@@ -106,13 +106,14 @@ func Parse(fs *flag.FlagSet, args []string, options ...Option) error {
...
-	configFileFlagName string
-	configFileParser   ConfigFileParser
-	envVarPrefix       string
-	envVarNoPrefix     bool
-	envVarIgnoreCommas bool
-	ignoreUndefined    bool
+	configFile             string
+	configFileFlagName     string
+	configFileParser       ConfigFileParser
+	configFileAllowMissing bool
```
peterbourgon: Minor nit: I think this parses better as

```suggestion
	allowMissingConfigFile bool
```

parse.go
```diff
@@ -142,6 +143,13 @@ func WithConfigFileParser(p ConfigFileParser) Option {
 	}
 }
 
+// WithConfigFileAllowMissing tells parse to allow/disallow missing of config file
+func WithConfigFileAllowMissing(allow bool) Option {
```
peterbourgon: Similarly

```suggestion
func WithAllowMissingConfigFile(allow bool) Option {
```

parse.go
```diff
@@ -142,6 +143,13 @@ func WithConfigFileParser(p ConfigFileParser) Option {
 	}
 }
 
+// WithConfigFileAllowMissing tells parse to allow/disallow missing of config file
```
peterbourgon:
```suggestion
// WithAllowMissingConfigFIle will permit parse to succeed, even if a provided 
// config file doesn't exist.
```

parse.go
Dragomir-Ivanov: Done

parse.go
Dragomir-Ivanov: Done

peterbourgon:
Nice! Thanks!

PR #48, opened 2019-12-02
Add support for TOML tables to fftoml package

GeorgeMac:
This adds support for TOML tables. It does so by concatenating nested keys with the `-` character.

For example:

```toml
[a]
b = "c"
```
Will set the flag `-a-b` with the value `"c"`.

```go
var b string
flagset.StringVar(&b, "a-b", "a string")
// ...
```

This swaps to a different toml parser (namely `github.com/pelletier/go-toml`), however, upon reflection the BurntSushi version is probably still useable. I can swap to attempt that if that would be more preferrable.

### DIFF
--- a/fftoml/fftoml.go
+++ b/fftoml/fftoml.go
@@ -18,15 +18,15 @@ func Parser(r io.Reader, set func(name, value string) error) error {
 
 // ConfigFileParser is a parser for the TOML file format. Flags and their values
 // are read from the key/value pairs defined in the config file.
-// Nested tables and keys are concatenated with a delimeter to derive the
+// Nested tables and keys are concatenated with a delimiter to derive the
 // relevant flag name.
 type ConfigFileParser struct {
-	delimeter string
+	delimiter string
 }
 
 // New constructs and configures a ConfigFileParser using the provided options.
 func New(opts ...Option) (c ConfigFileParser) {
-	c.delimeter = "."
+	c.delimiter = "."
 	for _, opt := range opts {
 		opt(&c)
 	}
@@ -41,31 +41,39 @@ func (c ConfigFileParser) Parse(r io.Reader, set func(name, value string) error)
 		return ParseError{Inner: err}
 	}
 
-	return parseTree(tree, "", c.delimeter, set)
+	return parseTree(tree, "", c.delimiter, set)
 }
 
 // Option is a function which changes the behavior of the TOML config file parser.
 type Option func(*ConfigFileParser)
 
-// TableDelimeter is an option which configures a delimeter
+// WithTableDelimiter is an option which configures a delimiter
 // used to prefix table names onto keys when constructing
 // their associated flag name.
-// The default delimeter is "."
-func TableDelimeter(d string) Option {
+// The default delimiter is "."
+//
+// For example, given the following TOML
+//
+//     [section.subsection]
+//     value = 10
+//
+// Parse will match to a flag with the name `-section.subsection.value` by default.
+// If the delimiter is "-", Parse will match to `-section-subsection-value` instead.
+func WithTableDelimiter(d string) Option {
 	return func(c *ConfigFileParser) {
-		c.delimeter = d
+		c.delimiter = d
 	}
 }
 
-func parseTree(tree *toml.Tree, parent, delimeter string, set func(name, value string) error) error {
+func parseTree(tree *toml.Tree, parent, delimiter string, set func(name, value string) error) error {
 	for _, key := range tree.Keys() {
 		name := key
 		if parent != "" {
-			name = parent + delimeter + key
+			name = parent + delimiter + key
 		}
 		switch t := tree.Get(key).(type) {
 		case *toml.Tree:
-			if err := parseTree(t, name, delimeter, set); err != nil {
+			if err := parseTree(t, name, delimiter, set); err != nil {
 				return err
 			}
 		case interface{}:
--- a/fftoml/fftoml.go
+++ b/fftoml/fftoml.go
@@ -6,26 +6,40 @@ import (
 	"io"
 	"strconv"
 
-	"github.com/BurntSushi/toml"
+	"github.com/pelletier/go-toml"
 	"github.com/peterbourgon/ff"
 )
 
 // Parser is a parser for TOML file format. Flags and their values are read
 // from the key/value pairs defined in the config file.
 func Parser(r io.Reader, set func(name, value string) error) error {
-	var m map[string]interface{}
-	_, err := toml.DecodeReader(r, &m)
+	tree, err := toml.LoadReader(r)
 	if err != nil {
 		return ParseError{Inner: err}
 	}
-	for key, val := range m {
-		values, err := valsToStrs(val)
-		if err != nil {
-			return ParseError{Inner: err}
+	return parseTree(tree, "", set)
+}
+
+func parseTree(tree *toml.Tree, parent string, set func(name, value string) error) error {
+	for _, key := range tree.Keys() {
+		name := key
+		if parent != "" {
+			name = parent + "-" + key
 		}
-		for _, value := range values {
-			if err = set(key, value); err != nil {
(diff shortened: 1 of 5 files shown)


ChrisHines:
Consider using a `.` (period) separator instead of `-` for nested keys. This would parallel the way the go tool passes args to the testing package (see https://golang.org/cmd/go/#hdr-Testing_flags) and also has precedence in the prior art as implemented by https://godoc.org/github.com/cespare/flagconf#Parse.

GeorgeMac:
I'm totally open to `.` have any reservations on that @peterbourgon ?

peterbourgon:
I personally never use `.` in my flag names, but the precedent exists, so I'd love to have it as an option.

GeorgeMac:
@peterbourgon I could have it attempt both? e.g. try `.` and if that fails try `_`.
I believe `set()` returns a sentinel error to denote flag does not exist, is that correct?

peterbourgon:
Interesting. That could definitely work, but it feels a bit too... implicit for my taste. I think I'd rather let users set the separator rune (string?) with a default of `.`.

GeorgeMac:
Perhaps an option on the parser itself?

e.g.
```go
package fftoml

func Parser(r io.Reader, set(name, value string) error) error {
    return ParserWith()(r, set)
}

type config struct {
  separator string
}

func FlagSeparator(s string) Option {
    return func(c *config) {
        c.separator = s
    }
}

func ParserWith(opts ...Option) ff.ConfigFileParser {
    config := config{separator: "."}
    for _, opt := range opts {
        opt(&config)
    }

    // ...
}
```

Then you could do `fftoml.ParserWith(fftoml.FlagSeparator("_"))` ?

peterbourgon:
Yeah, that's essentially what I had in mind 👍 unless you can think of a nicer API. (A couple of small nits with the example code, but that's better left for review.)

GeorgeMac:
@peterbourgon 👍 I like it. I pushed up that change.

fftoml/fftoml.go
```diff
@@ -6,27 +6,69 @@ import (
...
 // from the key/value pairs defined in the config file.
 func Parser(r io.Reader, set func(name, value string) error) error {
-	var m map[string]interface{}
-	_, err := toml.DecodeReader(r, &m)
-	if err != nil {
-		return ParseError{Inner: err}
+	return ParserWith()(r, set)
+}
+
+type config struct {
```
peterbourgon: I'd rather see this type as exported, so Config, to avoid having functions in the package's API that operate on unexported types

fftoml/fftoml.go
```diff
@@ -6,27 +6,69 @@ import (
...
+	separator string
+}
+
+// Option is a function which configures a fftoml.ConfigFileParser
```
peterbourgon: Please make sure all prefixing doc comments are full sentences with punctuation. In this case, add a period :)

```suggestion
// Option is a function which changes the behavior of the TOML config file parser.
```

fftoml/fftoml.go
```diff
@@ -6,27 +6,69 @@ import (
...
+type Option func(*config)
+
+// FlagSeparator is an option which configures a separator
+// to use when constructing a flag name
+func FlagSeparator(s string) Option {
```
peterbourgon: Hmm, I think this name is too vague. Is there a name for the TOML blocking feature, the thing we're segmenting config information with? I'd prefer something like... BlockSeparator?

fftoml/fftoml.go
```diff
@@ -6,27 +6,69 @@ import (
...
+
+type config struct {
+	separator string
+}
+
+// Option is a function which configures a fftoml.ConfigFileParser
+type Option func(*config)
+
+// FlagSeparator is an option which configures a separator
+// to use when constructing a flag name
```
peterbourgon: Please make this comment a bit more descriptive, and add a sentence like "The default separator is a period `.`."

fftoml/fftoml.go
```diff
@@ -6,27 +6,69 @@ import (
...
-	for key, val := range m {
-		values, err := valsToStrs(val)
+}
+
+// ParserWith configures and returns a new ConfigFileParser using the
+// provided slice of Option types
+// By default the returned Parser uses a "." as a separator
+func ParserWith(opts ...Option) ff.ConfigFileParser {
+	return func(r io.Reader, set func(name, value string) error) error {
+		config := config{separator: "."}
```
peterbourgon: Hmm, what do you think about adding a Parse method to the Config struct, matching the ConfigFileParser signature, and returning that? e.g.

```go
func ParserWith(opts ...Option) ff.ConfigFileParser {
    c := Config{
        separator: ".",
    }
    for _, opt := range opts {
        opt(&c)
    }
    return c.Parse
}
```

fftoml/fftoml_test.go
```diff
@@ -43,3 +52,62 @@ func TestParser(t *testing.T) {
...
+				fs    = flag.NewFlagSet("fftest", flag.ContinueOnError)
+			)
+
+			fs.StringVar(&found.String, testcase.stringKey, "", "string")
+			fs.Float64Var(&found.Float, testcase.floatKey, 0, "float64")
+			fs.Var(&found.Strings, testcase.stringsKey, "string slice")
+
+			if err := ff.Parse(fs, []string{},
+				ff.WithConfigFile("testdata/table.toml"),
+				ff.WithConfigFileParser(fftoml.ParserWith(testcase.opts...))); err != nil {
```
peterbourgon:
```suggestion
				ff.WithConfigFileParser(fftoml.ParserWith(testcase.opts...)),
			); err != nil {
```

fftoml/fftoml.go
ChrisHines: I believe the correct TOML term is a "table" https://github.com/toml-lang/toml#table

fftoml/fftoml.go
GeorgeMac: That is nice. Does it continue to make sense to call in Config? Perhaps there is a more suitable name. Like `ConfigFileParser` to match the `ff.ConfigFileParser` function signature?

fftoml/fftoml.go
GeorgeMac: How about `TableDelimiter(string)` ?

GeorgeMac:
@peterbourgon I made some adjustments based on your comments :) What do you think?

fftoml/fftoml.go
```diff
@@ -6,27 +6,78 @@ import (
...
+}
+
+// Option is a function which changes the behavior of the TOML config file parser.
+type Option func(*ConfigFileParser)
+
+// TableDelimeter is an option which configures a delimeter
+// used to prefix table names onto keys when constructing
+// their associated flag name.
+// The default delimeter is "."
+func TableDelimeter(d string) Option {
```
peterbourgon: Cool! Just a minor nit for consistency :)

```suggestion
func WithTableDelimeter(d string) Option {
```

fftoml/fftoml.go
```diff
@@ -6,27 +6,78 @@ import (
...
+	return parseTree(tree, "", c.delimeter, set)
+}
+
+// Option is a function which changes the behavior of the TOML config file parser.
+type Option func(*ConfigFileParser)
+
+// TableDelimeter is an option which configures a delimeter
+// used to prefix table names onto keys when constructing
+// their associated flag name.
+// The default delimeter is "."
```
peterbourgon:
```suggestion
// The default delimeter is "."
//
// For example, given the following TOML
//
//     [section.subsection]
//     value = 10
//
// Parse will match to a flag with the name `-section.subsection.value` by default.
// If the delimiter is "-", Parse will match to `-section-subsection-value` instead.
```

peterbourgon:
Oh, oops — it's spelled `Delimiter`, not `Delimeter` :)

GeorgeMac:
Ahh good catch 🤦‍♂. Cheers @peterbourgon !

PR #50, opened 2019-12-19
PlainParser: string without value currently becomes 'true'

peterbourgon:
It should not do this.


peterbourgon:
It seems reasonably difficult to change this behavior.

peterbourgon:
WONTFIX

PR #60, opened 2020-04-07
Allow WithConfigFileFlag and WithConfigFile to be meaningfully used together

jtolds:
Prior to this commit, WithConfigFileFlag and WithConfigFile could not be used together, as WithConfigFile would always take precedence.

It's conceivable that a tool author would want a tool user to be able to specify a config file if wanted, but fallback to a default if one was not provided. This change in semantics allows for this behavior. It's unlikely people used and wanted WithConfigFileFlag and WithConfigFile together prior to this change, so it's reasonably safe to add.

### DIFF
--- a/parse.go
+++ b/parse.go
@@ -67,7 +67,7 @@ func Parse(fs *flag.FlagSet, args []string, options ...Option) error {
 	})
 
 	// Third priority: config file (host).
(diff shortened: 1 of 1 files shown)


jtolds:
Nvm, https://github.com/peterbourgon/ff/pull/61

PR #61, opened 2020-04-07
Expand WithConfigFile/WithConfigFileFlag docs

jtolds:
Originally I wrote https://github.com/peterbourgon/ff/pull/60, but then after adding that patch and trying to use it, I figured out https://github.com/peterbourgon/ff/pull/60 was wrong and I just didn't know how to use the existing code. This PR adds the documentation I could have used to avoid thinking I needed https://github.com/peterbourgon/ff/pull/60.

### DIFF
--- a/parse.go
+++ b/parse.go
@@ -132,12 +132,8 @@ type Option func(*Context)
 
 // WithConfigFile tells Parse to read the provided filename as a config file.
 // Requires WithConfigFileParser, and overrides WithConfigFileFlag.
-//
-// If you want both a config file flag and a default value when one is
-// unprovided, use WithConfigFileFlag and set your chosen default when
-// defining the flag. Because config files should generally be
-// user-specifiable, this option should be rarely used. Prefer
-// WithConfigFileFlag.
+// Because config files should generally be user-specifiable, this option
+// should be rarely used. Prefer WithConfigFileFlag.
 func WithConfigFile(filename string) Option {
 	return func(c *Context) {
 		c.configFile = filename
--- a/parse.go
+++ b/parse.go
@@ -131,10 +131,13 @@ type Context struct {
 type Option func(*Context)
 
 // WithConfigFile tells Parse to read the provided filename as a config file.
-// Requires WithConfigFileParser, and overrides WithConfigFileFlag. If you
-// want both a config file flag and a default value when one is unprovided,
-// use WithConfigFileFlag and set your chosen default when defining the
-// flag.
+// Requires WithConfigFileParser, and overrides WithConfigFileFlag.
+//
+// If you want both a config file flag and a default value when one is
+// unprovided, use WithConfigFileFlag and set your chosen default when
+// defining the flag. Because config files should generally be
+// user-specifiable, this option should be rarely used. Prefer
+// WithConfigFileFlag.
 func WithConfigFile(filename string) Option {
 	return func(c *Context) {
 		c.configFile = filename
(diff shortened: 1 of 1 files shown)


parse.go
```diff
@@ -131,16 +131,20 @@ type Context struct {
...
 		c.configFile = filename
 	}
 }
 
 // WithConfigFileFlag tells Parse to treat the flag with the given name as a
-// config file. Requires WithConfigFileParser, and is overridden by
-// WithConfigFile.
+// config file. The caller will still need to define the flag in a FlagSet
+// (and can provide a default via that definition). Requires
+// WithConfigFileParser, and is overridden by WithConfigFile.
```
peterbourgon, line 147: I'm not sure how the additional text here provides any additional information. It is pretty clear to me that "treat the flag with the given name as a config file" necessarily implies that "the caller will need to define [that] fiag in [the] FlagSet and can provide a default via that definition".

parse.go
```diff
@@ -131,16 +131,20 @@ type Context struct {
 type Option func(*Context)
 
 // WithConfigFile tells Parse to read the provided filename as a config file.
-// Requires WithConfigFileParser, and overrides WithConfigFileFlag.
+// Requires WithConfigFileParser, and overrides WithConfigFileFlag. If you
+// want both a config file flag and a default value when one is unprovided,
+// use WithConfigFileFlag and set your chosen default when defining the
+// flag.
```
peterbourgon, line 137: 👍

parse.go
peterbourgon: Would it be sufficient to commit the other docs addition but not this one?

parse.go
peterbourgon: I would even maybe add another sentence somewhere, like

> Because config files should generally be user-specifiable, this option should be rarely used in practice. Prefer WithConfigFileFlag.

parse.go
peterbourgon: Or maybe an additional sentence like

> To specify a default config file, provide it as the default value of the corresponding flag — and consider also using the WithAllowMissingConfigFile option.

jtolds: > It is pretty clear to me that "treat the flag with the given name as a config file" necessarily implies that "the caller will need to define [that] fiag in [the] FlagSet and can provide a default via that definition".
I'm coming from a different library that has a similar feature... but defines the config flag for you. In that mindset, I totally expected ff to create its own flag and I wondered what the help text would be.

The proposed sentence is great, I'll swap to that.

parse.go
jtolds: A+

jtolds:
Great little library by the way! Exactly what I was looking for.

parse.go
```diff
@@ -132,6 +132,12 @@ type Option func(*Context)
 
 // WithConfigFile tells Parse to read the provided filename as a config file.
 // Requires WithConfigFileParser, and overrides WithConfigFileFlag.
+//
+// If you want both a config file flag and a default value when one is
+// unprovided, use WithConfigFileFlag and set your chosen default when
+// defining the flag. Because config files should generally be
+// user-specifiable, this option should be rarely used. Prefer
+// WithConfigFileFlag.
```
peterbourgon, line 140: This extra paragraph stutters. I'd prefer the comment block to read simply

```
// WithConfigFile tells Parse to read the provided filename as a config file.
// Requires WithConfigFileParser, and overrides WithConfigFileFlag.
// Because config files should generally be user-specifiable, this option
// should rarely be used. Prefer WithConfigFileFlag.
```

jtolds: sorry to make you spell it out for me, heh

fixed

PR #62, opened 2020-05-04
WIP: Define FlagSet interface

ebati:
FlagSet methods have the same function signatures as flag.FlagSet, so
*flag.FlagSet can be used as FlagSet, for different flag libraries one
can implement this interface. As an example AdaptPflag function converts
*pflag.FlagSet to FlagSet.

### DIFF
--- a/fftest/vars.go
+++ b/fftest/vars.go
@@ -7,6 +7,8 @@ import (
 	"reflect"
 	"strings"
 	"time"
+
+	"github.com/spf13/pflag"
 )
 
 // Pair returns a predefined flag set, and a predefined set of variables that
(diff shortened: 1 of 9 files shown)


ebati:
I updated the PR, added test with `pflag.FlagSet` but did not modify `ffcli`. Should `ffcli` also need to allow multiple FlagSet implementations?

peterbourgon:
> Should `ffcli` also need to allow multiple FlagSet implementations?

I guess it should be as simple as swapping `*flag.FlagSet` for `ff.FlagSet`, right? If so, yes please 👍

ebati:
The usage func is set in parse (`c.FlagSet.Usage = func (){...}`) so I couldn't swap types. 

Functions below are also not in FlagSet definition but we can add them:
- FlagSet.Args()
- FlagSet.Output()

PR #64, opened 2020-05-17
Support pflag.FlagSet

peterbourgon:
This PR refactors @ebati PR #62 in a few ways. It moves the pflag-specific code to its own package ffpflag, in order to isolate the dependency. Also, it isolates pflag-related tests to package ffpflag, in part to keep those tests single-responsibility-ish, and also because pflag has additional features that we want to include in the tests. Finally, it adds necessary docs.

@ebati, does this look OK to you?


peterbourgon:
Note that I still need to do some work to get this compatible with package ffcli.

parse.go
```diff
@@ -9,10 +9,24 @@ import (
 	"strings"
 )
 
+// FlagSet interface defines required minimum set of functions required for
+// parsing parameters.
+//
+// The standard library flag.FlagSet implements this interface. FromPflag
```
ebati, line 15: `FromPflag` is replaced with `ffpflag.NewFlagSet` so this needs to be updated.

ebati:
Before merging this PR, I think updating ffcli would be great. As we will have the opportunity to change interface definition if necessary.

peterbourgon:
I may still do this, but by another approach.

PR #65, opened 2020-05-30
parse: add env var name to usage text

edev0:
Thanks for this great library !

This PR automatically adds environment variable names to each flag's usage text.

For example, a definition such as:
```go
var (
    port  = fs.Int("port", 8080, "listen port for server")
    debug = fs.Bool("debug", false, "log debug information")
)
ff.Parse(fs, os.Args[1:], ff.WithEnvVarNoPrefix())
```
would result in the following usage text:
```
Usage of myservice:
  -debug
    	log debug information (also via env var DEBUG)
  -port int
    	listen port for server (also via env var PORT) (default 8080)
```

Also fixed some small typos.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,8 +98,8 @@ import (
 func main() {
 	fs := flag.NewFlagSet("myservice", flag.ExitOnError)
 	var (
-		port  = fs.Int("port", 8080, "listen port for server (also via PORT)")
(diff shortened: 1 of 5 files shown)


peterbourgon:
Sorry, I don't want this package to mutate the provided FlagSet.

edev0:
Understood, thanks !

edev0:
Understood, thanks !

PR #67, opened 2020-09-16
fix(ffcli): support sharing a flag with different default values

### LINKED ISSUE #68: Support async flag definition
I was thinking about various solutions to support my issue (#67), and one of them can be 100% managed thanks to ff, and I think it can also have other advantages for the ff users:

what do you think about supporting an alternative way to define flags, something like this maybe:

```go
fooFs := flag.NewFlagSet()
fooFs.Var(...)
barFs := flag.NewFlagSet()
barFs.Var(...)
root := &ffcli.Command{
	Subcommands: []*ffcli.Command{
		{Name: "foo", FlagSet: fooFs, Exec: noopHandler},
		{Name: "bar", FlagSet: barFs, Exec: noopHandler},
	},
	Exec: noopHandler,
}
```

becomes

```go
fooFsBuilder := func() (*flag.FlagSet, error) {
	fs := flag.NewFlagSet()
	fs.Var(...)
	return fs, nil
}
barFsBuilder := func() (*flag.FlagSet, error) {
	fs.NewFlagSet()
	fs.Var(...)
	return fs, nil
}
root := &ffcli.Command{
	Subcommands: []*ffcli.Command{
		{Name: "foo", FlagSetBuilder: fooFsBuilder, Exec: noopHandler},
		{Name: "bar", FlagSetBuilder: barFsBuilder, Exec: noopHandler},
	},
	Exec: noopHandler,
}
```

---

It can completely fix my issue and allow new workflows, i.e., if one of your flag default value requires some computing, i.e.

```go
ipAddress, _ := retrieveInternetIPAddress() // something that can fail, take time, or even lock
fs.StringVar(&ipAddress, "addr", ipAddress, "")
```

Thanks to this, a `program -h` will always be fast, even if a subcommand requires some computing for initializing flags.

I understand that it has one major drawback -> having two different ways of doing the same thing, but I think that it's totally understandable for people discovering the package, for me it's like the current `ParseAndBuild` doing `Parse`, then `Build`

I can start making a PR, just tell me what you think before I start please :)

moul:
~_disclaimer: hi, this PR only contains a test case to reproduce a bug, but no fix yet_~

~can someone help me understand how to fix it so I can finish the PR?~

thanks

*edit*: to support various edge cases, I added a `Command.FlagSetBuilder` optional callback

fixes #68

### DIFF
--- a/ffcli/command.go
+++ b/ffcli/command.go
@@ -8,7 +8,7 @@ import (
 	"strings"
 	"text/tabwriter"
 
-	"github.com/peterbourgon/ff/v3"
+	ff "github.com/peterbourgon/ff/v3"
 )
 
 // Command combines a main function with a flag.FlagSet, and zero or more
@@ -54,6 +54,10 @@ type Command struct {
 	// that the -h flag works as expected.
 	FlagSet *flag.FlagSet
 
+	// FlagSetBuilder is a callback used to configure the Command.FlagSet
+	// asynchronously, and only if needed.
+	FlagSetBuilder func() (*flag.FlagSet, error)
+
 	// Options provided to ff.Parse when parsing arguments for this command.
 	// Optional.
 	Options []ff.Option
@@ -90,6 +94,14 @@ type Command struct {
 func (c *Command) Parse(args []string) error {
 	if c.selected != nil {
 		return nil
+	}
+
+	if c.FlagSetBuilder != nil {
+		var err error
+		c.FlagSet, err = c.FlagSetBuilder()
+		if err != nil {
+			return err
+		}
 	}
 
 	if c.FlagSet == nil {
--- a/ffcli/command_test.go
+++ b/ffcli/command_test.go
@@ -439,6 +439,83 @@ func TestIssue57(t *testing.T) {
 	}
 }
 
+func TestSharedFlags(t *testing.T) {
+	for _, testcase := range []struct {
+		name             string
+		args             []string
+		sharedFlag       string
+		fooBuilderCalled bool
+		barBuilderCalled bool
+	}{
+		{
+			name:             "foo",
+			args:             []string{"foo"},
+			sharedFlag:       "foo-default",
+			fooBuilderCalled: true,
+		},
+		{
+			name:             "bar",
+			args:             []string{"bar"},
+			sharedFlag:       "bar-default",
+			barBuilderCalled: true,
+		},
+		{
+			name:             "foo -baz=test",
+			args:             []string{"foo", "-baz=test"},
+			sharedFlag:       "test",
+			fooBuilderCalled: true,
+		},
+		{
+			name:             "bar -baz=test",
+			args:             []string{"bar", "-baz=test"},
+			sharedFlag:       "test",
+			barBuilderCalled: true,
+		},
+		{
+			name:             "foo -baz=",
+			args:             []string{"foo", "-baz="},
+			sharedFlag:       "",
+			fooBuilderCalled: true,
+		},
+		{
+			name:             "bar -baz=",
+			args:             []string{"bar", "-baz="},
+			sharedFlag:       "",
+			barBuilderCalled: true,
+		},
+	} {
+		t.Run(testcase.name, func(t *testing.T) {
+			var fooBuilderCalled, barBuilderCalled bool
+			sharedFlag := "init-value"
+			fooFsBuilder := func() (*flag.FlagSet, error) {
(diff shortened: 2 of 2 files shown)


ffcli/command.go
```diff
@@ -8,7 +8,7 @@ import (
 	"strings"
 	"text/tabwriter"
 
-	"github.com/peterbourgon/ff/v3"
+	ff "github.com/peterbourgon/ff/v3"
```
moul, line 11: my current goimports setup requires this, tell me if it's a problem so I can edit this file without IDE hook and come back to the previous state

ffcli/command_test.go
```diff
@@ -439,6 +439,65 @@ func TestIssue57(t *testing.T) {
...
+func TestSharedFlags(t *testing.T) {
+	for _, testcase := range []struct {
+		name       string
+		args       []string
+		sharedFlag string
+	}{
+		{
+			name:       "foo",
+			args:       []string{"foo"},
+			sharedFlag: "foo-default",
```
moul, line 451: This testcase fails, when printing usage, we can correctly see `foo-default`, but when you run the foo command without CLI flags, then you get `bar-default`.

I tried to add some printf-debugging in the library and I can't find why the pointer is modified because, the `ff.Parse` function is only called against the root's and foo's `ffcli.Command.FlagSet`s

Any idea what's the root cause and how to make a workaround?

ffcli/command_test.go
```diff
@@ -439,6 +439,65 @@ func TestIssue57(t *testing.T) {
...
+			args:       []string{"bar", "-baz="},
+			sharedFlag: "",
+		},
+	} {
+		t.Run(testcase.name, func(t *testing.T) {
+			sharedFlag := "init-value"
+			fooFs := flag.NewFlagSet("foo", flag.ContinueOnError)
+			barFs := flag.NewFlagSet("bar", flag.ContinueOnError)
+			fooFs.StringVar(&sharedFlag, "baz", "foo-default", "testing")
+			barFs.StringVar(&sharedFlag, "baz", "bar-default", "testing")
```
peterbourgon, line 484: Haha! This is a bug (?) with stdlib flag.FlagSet. 

The default value provided to e.g. fs.StringVar is assigned to the associated variable _at declaration time_, i.e. when you call fs.StringVar, and _not_ when you call fs.Parse. So the default value you see will be whichever value is given in the final fs.StringVar call you make — in your test, it's this one.

I think the only fix here would be, somehow, avoiding assigning a default value in the declaration, and populating it later. Or defining a separate e.g. DeferredFlagVar type that encapsulated this behavior? I don't know offhand what's best. Ugh.

moul: I can confirm it's not related with ff :)

```go
sharedFlag := "init-value"
fooFs := flag.NewFlagSet("foo", flag.ContinueOnError)
barFs := flag.NewFlagSet("bar", flag.ContinueOnError)
fooFs.StringVar(&sharedFlag, "baz", "foo-default", "testing")
barFs.StringVar(&sharedFlag, "baz", "bar-default", "testing")
_ = fooFs.Parse([]string{""})
fmt.Println(*sharedFlag) // bar-default
```

I'll ask on golang-nuts if I need to report a bug and see if I can make a contribution on the go codebase :)

moul:
not related with this repo, closing

sorry for the noise

moul:
https://groups.google.com/g/golang-nuts/c/n5iiO6UY-eU/m/jlfpykFsBwAJ

moul:
@peterbourgon I just updated this PR to support what I proposed in #68

peterbourgon:
OK, I understand what you're going for here, and I'm not against it as a feature, but we'd need to find a less... awkward? implementation. The package API is already really large (IMO) and I'm basically against adding an entirely new field to the Command type just for this rather esoteric use case. I'll think about it and see if there's a different way...

moul:
Hey @peterbourgon, did you found a better approach?

peterbourgon:
I think the only viable approach here is to do the work at the variable level, rather than at the FlagSet level. It would probably need to be an implementation of flag.Value with some kind of late binding behavior, or maybe a type that managed multiple flag.Values, one per flag set, and had some method to resolve them to a single "shared" value with the correct default. But the way the stdlib package flag is implemented makes this very difficult, I haven't figured anything out just yet.

peterbourgon:
[Something like this](https://play.golang.org/p/unU2Xm_3m-G).

PR #71, opened 2020-11-10
ff: add WithConfigFileVia option

### LINKED ISSUE #70: Not sure how to properly to share config file between root flag and subcommands.
I want something like this to work: `mycmd -config configfile subcommand -subarg ...` where one config file has values for both root and any subcommands. 

Having one config file per subcommand with it's own `-config` argument is too messy and user unfriendly to consider, the one root level `-config` should be enough and the config file should preferably not have to be reread multiple times.

I'm not sure how to express this in a clean way using ffcli.

thomasf:
as discussed in #70 

closes #70

### DIFF
--- a/parse.go
+++ b/parse.go
@@ -66,15 +66,20 @@ func Parse(fs *flag.FlagSet, args []string, options ...Option) error {
 		provided[f.Name] = true
 	})
 
+	var configFile string
+	if c.configFileVia != nil {
+		configFile = *c.configFileVia
+	}
+
 	// Third priority: config file (host).
-	if c.configFile == "" && c.configFileFlagName != "" {
+	if configFile == "" && c.configFileFlagName != "" {
 		if f := fs.Lookup(c.configFileFlagName); f != nil {
-			c.configFile = f.Value.String()
+			configFile = f.Value.String()
 		}
 	}
 
-	if parseConfig := c.configFile != "" && c.configFileParser != nil; parseConfig {
-		f, err := os.Open(c.configFile)
+	if parseConfig := configFile != "" && c.configFileParser != nil; parseConfig {
+		f, err := os.Open(configFile)
 		switch {
 		case err == nil:
 			defer f.Close()
@@ -117,7 +122,7 @@ func Parse(fs *flag.FlagSet, args []string, options ...Option) error {
 
 // Context contains private fields used during parsing.
 type Context struct {
-	configFile             string
+	configFileVia          *string
 	configFileFlagName     string
 	configFileParser       ConfigFileParser
 	allowMissingConfigFile bool
@@ -135,8 +140,16 @@ type Option func(*Context)
 // Because config files should generally be user-specifiable, this option
 // should be rarely used. Prefer WithConfigFileFlag.
 func WithConfigFile(filename string) Option {
+	return WithConfigFileVia(&filename)
+}
+
+// WithConfigFileVia tells Parse to read the provided filename as a config file.
+// Requires WithConfigFileParser, and overrides WithConfigFileFlag.
+// This is useful for sharing a single root level flag for config files among
+// multiple ffcli subcommands.
+func WithConfigFileVia(filename *string) Option {
 	return func(c *Context) {
-		c.configFile = filename
+		c.configFileVia = filename
 	}
 }
 
--- a/parse_test.go
+++ b/parse_test.go
@@ -1,11 +1,14 @@
 package ff_test
 
 import (
+	"context"
+	"flag"
 	"os"
 	"testing"
 	"time"
 
 	"github.com/peterbourgon/ff/v3"
+	"github.com/peterbourgon/ff/v3/ffcli"
 	"github.com/peterbourgon/ff/v3/fftest"
 )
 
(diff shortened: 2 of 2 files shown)


thomasf:
I see a bug. will amend

thomasf:
WithConfigFileVia could maybe be named WithConfigFileVar because it is a similar pattern to how FlagSet.StringVar(...) is named, not entirely the same thing though.

parse.go
```diff
@@ -118,6 +136,7 @@ func Parse(fs *flag.FlagSet, args []string, options ...Option) error {
 // Context contains private fields used during parsing.
 type Context struct {
 	configFile             string
```
peterbourgon, line 138: I think you can refactor WithConfigFile to be sugar for WithConfigFileVia, which would let you remove this field and leave Parse essentially untouched.

thomasf: done

thomasf:
Maybe the test is a bit misplaced (ffcli instead of ff package) but it does ensure that my particular use case works.

parse.go
```diff
@@ -67,14 +67,15 @@ func Parse(fs *flag.FlagSet, args []string, options ...Option) error {
 	})
 
 	// Third priority: config file (host).
-	if c.configFile == "" && c.configFileFlagName != "" {
+	if (c.configFileVia == nil || *c.configFileVia == "") && c.configFileFlagName != "" {
```
peterbourgon, line 70:
```suggestion
	var configFile string
	if c.configFileVia != nil {
		configFile = *c.configFileVia
	}
	if configFile == "" && c.configFileFlagName != "" {
```

parse.go
```diff
@@ -67,14 +67,15 @@ func Parse(fs *flag.FlagSet, args []string, options ...Option) error {
...
-			c.configFile = f.Value.String()
+			v := f.Value.String()
+			c.configFileVia = &v
 		}
 	}
 
-	if parseConfig := c.configFile != "" && c.configFileParser != nil; parseConfig {
-		f, err := os.Open(c.configFile)
+	if parseConfig := c.configFileVia != nil && *c.configFileVia != "" && c.configFileParser != nil; parseConfig {
+		f, err := os.Open(*c.configFileVia)
```
peterbourgon, line 78:
```suggestion
	if parseConfig := configFile != "" && c.configFileParser != nil; parseConfig {
		f, err := os.Open(configFile)
```

parse_test.go
```diff
@@ -254,3 +257,58 @@ func TestParseConfigFile(t *testing.T) {
...
+	t.Parallel()
+
+	rootFS := flag.NewFlagSet("root", flag.ContinueOnError)
+	config := rootFS.String("config-file", "", "")
+	iFlag := rootFS.Int("i", 0, "")
+	sFlag := rootFS.String("s", "", "")
+
+	subFS := flag.NewFlagSet("subcommand", flag.ContinueOnError)
+	dFlag := subFS.Duration("d", time.Second, "")
+	bFlag := subFS.Bool("b", false, "")
```
peterbourgon, line 271:
```suggestion
	var (
		rootFS = flag.NewFlagSet("root", flag.ContinueOnError)
		config = rootFS.String("config-file", "", "")
		i     = rootFS.Int("i", 0, "")
		s     = rootFS.String("s", "", "")

		subFS = flag.NewFlagSet("subcommand", flag.ContinueOnError)
		d     = subFS.Duration("d", time.Second, "")
		b     = subFS.Bool("b", false, "")
	)
```

parse_test.go
```diff
@@ -254,3 +257,58 @@ func TestParseConfigFile(t *testing.T) {
...
+		t.Error(*dFlag)
+	}
+	if *sFlag != "bar" {
+		t.Error(*sFlag)
+	}
+	if *iFlag != 99 {
+		t.Error(*iFlag)
+	}
+
+}
```
peterbourgon, line 314:
```suggestion
	if want, have := time.Hour, *dFlag; want != have {
		t.Errorf("d: want %v, have %v", want, have)
	}
	if want, have := false, *bFlag; want != have {
		t.Errorf("b: want %v, have %v", want, have)
	}
	if want, have := "bar", *sFlag; want != have {
		t.Errorf("s: want %q, have %q", want, have)
	}
	if want, have := 99, *iFlag; want != have {
		t.Errorf("i: want %d, have %d", want, have)
	}
}
```

thomasf: don't really see the need for this, IMO it just adds noise but it's your package so I've made the change anyway

peterbourgon: Well, it did allow me to catch and fix the bug below :)

```go
	if !*bFlag {
		t.Error(*dFlag)
	}
```

parse_test.go
```diff
@@ -1,11 +1,14 @@
...
 
 import (
+	"context"
+	"flag"
 	"os"
 	"testing"
 	"time"
 
 	"github.com/peterbourgon/ff/v3"
+	"github.com/peterbourgon/ff/v3/ffcli"
```
peterbourgon, line 11: :\ I'll refactor this later.

peterbourgon (approved):
Cool!

PR #73, opened 2020-11-28
Provide more contextualized usage messages

ChrisHines:
Usage messages should include the names and flags for all commands
from the root to the terminal command. Previously usage functions
could only access the terminal command's data, which made it difficult
to automatically generate a fully contextuatlized usage message. The
programmer was forced to manually copy the information into each
subcommands usage function.

This commit changes Command.Parse to collect the sequence of commands
found from the beginning of the parse to the terminal command found.
It also changes the signature of UsageFunc to accept the sequence of
commands found during parsing. These changes provide implementations
of UsageFunc with the data necessary to automatically generate more
detailed usage messages with less work required by programmers.

Note: The UsageFunc signature change is not backwards compatible and
requires increasing the major version of the module.

### DIFF
--- a/ffcli/command.go
+++ b/ffcli/command.go
@@ -47,7 +47,7 @@ type Command struct {
 	// command's short usage, short help, and long help strings, subcommands,
 	// and available flags. Optional; if not provided, a suitable, compact
 	// default is used.
-	UsageFunc func(c *Command) string
+	UsageFunc func(c ...*Command) string
 
 	// FlagSet associated with this command. Optional, but if none is provided,
 	// an empty FlagSet will be defined and attached during the parse phase, so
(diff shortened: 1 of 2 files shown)


peterbourgon:
> Usage messages should include the names and flags for all commands from the root to the terminal command.

I don't buy this, on the face of it, at least. Why?

peterbourgon:
An additional observation, just as a sort of window into my thoughts...

There is a design decision I made early on which I'm frequently re-assessing. Parse mutates the command tree, laying breadcrumbs in unexported fields so that Run knows where to go. Alternatively, Parse could leave the command tree un-modified, and return some kind of e.g. ParseContext, which would be an input to Run.

I didn't do it this way mostly because it would complect the API with an opaque type. It might also have tricky consequences with e.g. FlagSet parsing, I haven't fully thought that through. But it would make certain things easier, like re-using a command tree. And, maybe, things like this PR?

peterbourgon:
I hope this will be addressed by #113.

PR #77, opened 2021-03-18
Add Note for flag defaults when env var matches

gilcrest:
Hi @peterbourgon - thanks so much for this library. I found it at just the right time to save me a bunch of work! One thing that I had to do some testing around to make sure I understood how it worked was how ff would handle flag defaults. I like the way it works currently, I just thought I could add a note in to help newcomers like myself understand how the logic around precedence works for flag defaults. Does this make sense? Thanks for considering! Dan

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,6 +106,7 @@ func main() {
 	fmt.Printf("port %d, debug %v\n", *port, *debug)
 }
 ```
+> Note: Flag defaults are not considered when a matching environment variable is found. In the first example below, the default for the `port` *flag* is set to 8080, however, upon parsing with `ff.Parse` the `port` **variable** is actually set to 9090 as there was a `PORT` environment variable found. The `debug` **variable** in this example uses the default and is set to false as there is no matching environment variable to override it.
 
 ```
 $ env PORT=9090 myservice
--- a/README.md
+++ b/README.md
@@ -106,7 +106,7 @@ func main() {
 	fmt.Printf("port %d, debug %v\n", *port, *debug)
 }
 ```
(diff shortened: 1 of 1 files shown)


peterbourgon:
Glad you find it useful! I'm a little confused by your confusion, to be honest :) Did you expect the flag default to be applied when a matching env var was provided, in some circumstance?

gilcrest:
It is confusing... I guess, yeah, I wasn't sure how it worked exactly by reading the README. In the example you provide you set a default of 8080 for the flag. I thought, perhaps, since flags take precedence, your Parse logic would follow the normal flag rules, meaning that given a non-zero default value was provided, consider that first before checking the environment variable. I thought the logic might only check for environment variables in the case that there was no flag given as a command line argument and the default was a non-zero value. Essentially though, the default is always at the bottom of the logic hierarchy regardless of whether or not it's the type's zero value or not, correct?

I guess I'm the only one confused, but I had to run a set of tests to make sure it would work like this. I like the way it works, I just thought a little additional clarity couldn't hurt. Sorry to take your time!

peterbourgon:
No no, I'm just trying to understand :) I can definitely add something about this to the README, though I think I'll opt for something a bit shorter. Stay tuned...

gilcrest:
ok, cool - I'll see if I can take a stab at something a bit more terse tomorrow and you can take it or leave it! Thanks!

gilcrest:
Gave it a shot with a second commit. My apologies for the superfluity of words in the original - no worries if you prefer something different.

Thanks again for considering - hope you have a great weekend.

Best,

Dan

peterbourgon:
I done did it in #78. Thanks for the initiative!

Issue #81, opened 2021-06-17
Question Re: Yaml Parser


aramcodz:
On our team is using the [ff Yaml parser](https://github.com/peterbourgon/ff/blob/main/ffyaml/ffyaml.go). 

We came across a possible use case where we may have the config.yaml file that has several string/scalar fields that contain an empty str value, as in:
  
```
  propertyOne: "someVal"
  propertyTwo: 
  log-level: "Debug" 
```

where `propertyTwo` is  an empty string field.

When the yaml parser tries to read in one of these yaml files, it returns this error: 
  
`   error parsing YAML config: couldn't convert %!q(<nil>) (type <nil>) to string`
   
I noticed that the test case "no value" in https://github.com/peterbourgon/ff/blob/main/ffyaml/ffyaml_test.go actually expects an empty string value for a yaml field to return this error.

But, the YAML 1.2 Spec does say it allows empty scalar values that should resolve to the "null" tag (tag:yaml.org,2002:null). See https://yaml.org/spec/1.2/spec.html#id2805071. 

Is there a specific reason that ff Yaml Parser returns this (not very helpful/descriptive) error for an empty string value in the yaml file? We assumed that empty field values would be supported.

peterbourgon:
If it's not to spec I'll happily take a fix, but with the caveat that I'm no YAML guru, I'm not sure this can actually work. You call this an empty string value, but how do you know it's not a boolean? Or an array?

peterbourgon:
At a minimum the error message should be improved by annotating (prefixing) it with the key that triggered the problem.

aramcodz:
> If it's not to spec I'll happily take a fix, but with the caveat that I'm no YAML guru, I'm not sure this can actually work. You call this an empty string value, but how do you know it's not a boolean? Or an array?


From the Yaml 1.2 spec ([section 7.2 - Empty Nodes](https://yaml.org/spec/1.2/spec.html#id2786563)):

> YAML allows the node content to be omitted in many cases. Nodes with empty content are interpreted as if they were plain scalars with an empty value. Such nodes are commonly resolved to a “null” value.

I believe that for any scalar field where they key is present but has empty value, the YAML Spec itself, says that it interprets that value as a `null` "type".  See https://stackoverflow.com/a/64462925/5414775. 

So, perhaps I should be more specific in YAML terms. 

In my example here:

```
  propertyOne: "someVal"
  propertyTwo: 
  log-level: "Debug" 
```

[...]

PR #82, opened 2021-06-28
Proposed change to YAML parser functionality 

aramcodz:
YAML parser will now allowing empty values for each of the 6 supported golang datatypes ff lib (string, int, bool, duration, float, slice), the ff lib would set any "empty" value's flag to the golang "zero value" for that type. 

reference: https://github.com/peterbourgon/ff/issues/81

Signed-off-by: Aram Openden <Aram.Openden1@ibm.com>


ffyaml/ffyaml.go
```diff
@@ -1,4 +1,4 @@
-// Package ffyaml provides a YAML config file paser.
+// Package ffyaml provides a YAML config file parser.
```
peterbourgon, line 1: Thanks 😬

ffyaml/testdata/empty_basic_vals.yaml
```diff
@@ -0,0 +1,6 @@
+emptyStr:
+emptyInt:
+emptyBool:
+emptyDur:
+emptyFloat:
+emptySlice: []
```
peterbourgon, line 6: Newline please! :)

ffyaml/ffyaml.go
```diff
@@ -48,29 +55,36 @@ func valsToStrs(val interface{}) ([]string, error) {
...
-		return strconv.FormatFloat(v, 'g', -1, 64), nil
+		rtnStr = strconv.FormatFloat(v, 'g', -1, 64)
+	case nil:
+		return nil, nil
 	default:
-		return "", ff.StringConversionError{Value: val}
+		return nil, ff.StringConversionError{Value: val}
 	}
+
+	return &rtnStr, err
```
peterbourgon, line 87: Why can't you return from each case block?

ffyaml/ffyaml.go
```diff
@@ -48,29 +55,36 @@ func valsToStrs(val interface{}) ([]string, error) {
...
-		return strconv.Itoa(v), nil
+		rtnStr = strconv.Itoa(v)
 	case int64:
-		return strconv.FormatInt(v, 10), nil
+		rtnStr = strconv.FormatInt(v, 10)
 	case float64:
-		return strconv.FormatFloat(v, 'g', -1, 64), nil
+		rtnStr = strconv.FormatFloat(v, 'g', -1, 64)
+	case nil:
+		return nil, nil
```
peterbourgon, line 82: So if I'm reading this right, if the YAML parser yields a key with a nil value — presumably what happens when you just write `key: ` with nothing to the right of the colon — we would previously treat this as a StringConversionError, and are now special-casing it and treating it as if the `key` flag hadn't been provided at all. Is that correct?

ffyaml/ffyaml_test.go
```diff
@@ -71,3 +71,75 @@ func TestParser(t *testing.T) {
...
+	for _, testcase := range []struct {
+		name string
+		file string
+		want fftest.Vars
+	}{
+		{
+			name: "preset FS vals not overwritten",
+			file: "testdata/empty_basic_vals.yaml",
+			want: fftest.Vars{S: "EMPTY_DEFAULT", I: -500000, F: 42.42,
+				D: 86400 * time.Second, B: true, X: []string{"strVal1", "strVal2"}},
```
peterbourgon, line 128: Please fix this.

ffyaml/ffyaml_test.go
```diff
@@ -71,3 +71,75 @@ func TestParser(t *testing.T) {
 		})
 	}
 }
+
+func TestParserAllTypesEmptyVals(t *testing.T) {
```
peterbourgon, line 75: Why does this need to be a separate test, rather than part of TestParser?

ffyaml/ffyaml_test.go
```diff
@@ -71,3 +71,75 @@ func TestParser(t *testing.T) {
...
+func checkParseErr(t *testing.T, pError error) {
+	if pError != nil {
+		var errMsg string
+		errMsg = "ParseError: " + pError.Error()
+		if errors.Is(pError, ff.StringConversionError{}) {
+			errMsg = errMsg + ";errType: " + "StringConversionError"
+		}
+		t.Fatal(errMsg)
+	}
+}
```
peterbourgon, line 114: Nope 😉 Do it properly at the callsites.

```suggestion
```

ffyaml/ffyaml_test.go
```diff
@@ -71,3 +71,75 @@ func TestParser(t *testing.T) {
...
+
+func TestEmptyValsDontOverwritePresets(t *testing.T) {
```
peterbourgon, line 116: Why does this need to be a separate test, rather than part of TestParser?

fftest/vars.go
```diff
@@ -27,6 +27,42 @@ func Pair() (*flag.FlagSet, *Vars) {
 	return fs, &v
 }
 
+// EmptyValsFS returns a predefined flag set, and a predefined set of variables that
+// have been registered into it.
+// Custom Test Flagset creation function for the Empty (YAML) Values parsing, using more
+// specific flag names.
```
peterbourgon, line 33: Why is this necessary?

fftest/vars.go
```diff
@@ -27,6 +27,42 @@ func Pair() (*flag.FlagSet, *Vars) {
...
+	fs.DurationVar(&v.D, "emptyDur", 0*time.Second, "time.Duration")
+	fs.Var(&v.X, "emptySlice", "collection of strings (repeatable)")
+
+	return fs, &v
+}
+
+func PresetValsFS() (*flag.FlagSet, *Vars) {
+	fs := flag.NewFlagSet("ffpresettest", flag.ContinueOnError)
+
+	//Set Defaults to be something OTHER than GoLang defaults for each type
```
peterbourgon, line 51: Why is this necessary?

ffyaml/ffyaml_test.go
aramcodz: > Please fix this.

Fix what?  I am not clear on what you mean here.

ffyaml/ffyaml_test.go
aramcodz: Sorry, I don't understand what you mean by this: 

> Do it properly at the callsites.

ffyaml/ffyaml_test.go
peterbourgon: Sorry, that was ambiguous.

```suggestion
			want: fftest.Vars{
				S: "EMPTY_DEFAULT", 
				I: -500000, 
				F: 42.42,
				D: 86400 * time.Second, 
				B: true, 
				X: []string{"strVal1", "strVal2"},
			},
```

ffyaml/ffyaml_test.go
peterbourgon: Do not define or use a `checkParseErr` helper function. Check the error directly at the callsites.

ffyaml/ffyaml.go
aramcodz: > and are now special-casing it and treating it as if the `key` flag hadn't been provided at all.

Well, the difference here is that in `valToStr()` (lines 81-82), in the case of a nil `value` for the value part of the YAML entry, `valToStr()` simply returns a `nil` (str pointer).

Then on line 29 of Parser(), we do a nil-check of the pointer for that value. Only if it is Not-Nil,  do we then go ahead and update the  `value` variable in the `set()` function.

peterbourgon: Yes, so we're changing 

```yaml
flagname: 
```

from being an error, to being a no-op, taking no action on this line at all, and specifically _not_ treating it as equivalent to `-flagname=""`. Is that correct?

ffyaml/ffyaml.go
aramcodz: > Why can't you return from each case block?

I struggled with this very question for a bit. 

From what I could tell, using the `&` operator to convert the string var to a pointer was not allowed in the same (in-line) operation as the function calls to `string()`, line 68 or `strconv.Format*` as on lines 72,74, etc. At least, I saw an example that said it was not correct syntax.

ffyaml/ffyaml.go
aramcodz: > from being an error, to being a no-op, taking no action on this line at all, and specifically not treating it as equivalent to -flagname="". Is that correct?


@peterbourgon. Yes, that is correct.

fftest/vars.go
aramcodz: I wanted to ensure I was covering the "Negative-test scenario" that confirms that the No-Op for the Flagset update would not "overwrite" preset set of flags that are defined with something Other Than the Zero-Value for each type. 
  
So, this test fixture, `PresetValsFS`, in combination with test function `TestEmptyValsDontOverwritePresets()` ensure that test case is covered.

ffyaml/testdata/empty_basic_vals.yaml
aramcodz: Yup - Done. See latest push.

ffyaml/ffyaml_test.go
aramcodz: OK, got it now - Sure, Formatting has been updated to enhance readability/clarity.

See Latest.

ffyaml/ffyaml.go
peterbourgon: @aramcodz Great! Thanks for confirming. So I agree with you that

```yaml
foo: 123
bar: 
log-level: debug
```

is valid YAML, which declares the key bar as having an empty/null value. And I think I'm convinced that the current behavior of the YAML parser, which treats this as an error condition, is incorrect and should be changed. But I'm not (yet) convinced that the behavior in this PR is what we should be doing. My intuition is that the above config should not be interpreted as bar not being passed at all, e.g.

```
myprogram -foo="123"         -log-level="debug"
```

but should rather be interpreted as bar being passed and given an empty value, e.g.

```
myprogram -foo="123" -bar="" -log-level="debug"
```

Do you have an opinion? If you prefer the first form, why? 🙃

aramcodz: Yes, We prefer the first form (*Note: I've been working closely with my colleague, https://github.com/David-N-Perkins on this issue, and we conferred on the response).

We don't think that we can assume a blank YAML field value entry = an empty string value.  Let me explain:

1. Since the YAML specification treats a blank value as being a "null field" entry and not an empty string value, ("") we think that the ff library should respect that.
2.  _Not specifying a value_ is very different from specifying an empty value. A yaml config could have `bar: ""` to specify a blank (`string`) value.
3. Assuming that a blank YAML value equals `""` _also assumes a type_ (`string` type). We think `nil` is a much better representation of what was provided in the config _than assuming a type and a value._

Perhaps, ignoring a blank entry is not the best approach, but we see that as an issue with the existing ff lib design. A better approach might be **using pointers** instead of types for each config (FlagSet) value, so the user can differentiate between a config not being specified, specified without a value, or specified with a value. But we don't see how the current design allows for that.

Either way, we think _treating blanks as **not present**_ is better than assuming a blank YAML value is always an empty string.

ffyaml/ffyaml_test.go
aramcodz: > Why does this need to be a separate test, rather than part of TestParser?

If I use the `EmptyValsFS()` test fixture function to set up the initial test flagset (line 90):

`fs, testVars := fftest.EmptyValsFS()`

...then I think I need to have this test case as a separate test function, `TestParserAllTypesEmptyVals()`. 

I prefer to use this new test fixture function that I created [`EmptyValsFS()`], as it specifies that each type/field in the test field uses specific fields names (e.g.emptyStr, emptyInt, etc.).  I was intentionally creating this new test fixture to help self-document the specifics of this test case in combination with the test function (basically I wanted all the names to match).

fftest/vars.go
aramcodz: > Why is this necessary?

See my response here -> https://github.com/peterbourgon/ff/pull/82#discussion_r660864051.

It is not technically _necessary_. 

I chose to create `EmptyValsFS()` to use more descriptive field names in this function and in the `empty_basic_vals.yaml` for this test case. 

I thought that might be helpful on combination with function TestParserAllTypesEmptyVals() to more clearly self-document this new/additional functionality of the YAML parser that I am proposing in this PR.

ffyaml/ffyaml_test.go
aramcodz: Done. I have reversed the refactoring of Checking the error logic, "inline'ing" the previous checkParseErr() functionality.

ffyaml/ffyaml_test.go
aramcodz: > Why does this need to be a separate test, rather than part of TestParser?

This test function covers the "Negative-test scenario" that confirms that the No-Op for the Flagset update would not "overwrite" preset set of flags that are defined with something Other Than the Zero-Value for each type.

In order to do that, I needed to use a different test fixture method to set up the test, `PresetValsFS()`. This function sets the values of the Flagset variables to something other than the Zero value defaults for each type in golang. 

So on line 132, I call:

`fs, testVars := fftest.PresetValsFS()`

The test fixture, `PresetValsFS()`, in combination with test function `TestEmptyValsDontOverwritePresets()` ensure that the Negative-Test-Don't-Overwrite-Pre-defined-FlagSet-Values test case is covered.

aramcodz:
@peterbourgon: I believe I have now responded to all of your questions and suggested changes. Would you mind taking another look at the latest code changes in the PR and my responses, please?

ffyaml/ffyaml.go
peterbourgon: > Since the YAML specification treats a blank value as being a "null field" entry and not an empty string value, ("") we think that the ff library should respect that.

As far as I can see, "null field" is a concept defined by the YAML spec but without any specific semantics. When it needs to be interpreted in the context of a YAML AST being translated to flag values, it is not at all obvious to me what it should mean. 

Concretely, your PR makes

```yaml
foo: bar
baz:
```

equivalent to 

```yaml
foo: bar
```

I could be convinced that's the correct interpretation, but I don't believe it by default. Given `baz: `, it's much more intuitive to me that I get `-baz=""` out the other side, versus nothing at all. Why else would you put `baz: ` in the config in the first place? — I could be wrong! I don't know YAML super good, is there anything in the spec that would maybe push my belief one way or the other?

> Not specifying a value is very different from specifying an empty value. A yaml config could have bar: "" to specify a blank (string) value.

That's super confusing. To me it seems obvious that

```yaml
bar: ""
```

is, to a human, exactly the same as

```yaml
bar: 
```

In what [common] circumstances would someone leverage this null value property of YAML?

> Assuming that a blank YAML value equals "" also assumes a type (string type). We think nil is a much better representation of what was provided in the config than assuming a type and a value.

Given a flag `-key=val`, then yes, the val is by definition a string type. Maybe that's a disconnect here? All ff config files are by definition a map of strings to strings. If you want to specify a value as a non-string type, that's fine, but it will be converted to a string before it's received by ff.

>  A better approach might be using pointers instead of types for each config (FlagSet) value, so the user can differentiate between a config not being specified, specified without a value, or specified with a value. 

ff stands for flags-first, and it interprets everything, including parameters parsed from a config file, as if they were flags provided at the command line. The way to express "not specified" when enumerating flags at the command line is to not provide the flag. Therefore, intuitively, the way to express "not specified" when enumerating parameters in a config file is to not include the parameter in the file. 

Again — I'm open to being convinced otherwise. But show me something that will convince me! 🙃

ffyaml/ffyaml_test.go
```diff
@@ -1,12 +1,12 @@
...
 import (
-	"testing"
-	"time"
-
+	"errors"
 	"github.com/peterbourgon/ff/v3"
 	"github.com/peterbourgon/ff/v3/fftest"
 	"github.com/peterbourgon/ff/v3/ffyaml"
+	"testing"
+	"time"
```
peterbourgon, line 9: Please update your editor to run `goimports` on save (or whatever). Import blocks should have two sections: stdlib, newline, non-stdlib.

fftest/vars.go
peterbourgon: > I do not think having field names s, i, b, d, f in the test fixture function or in the empty_basic_values.yaml file contributes to that unerstanding.

I'm confused. There is no difference between the FlagSets returned by this function and Pair, except for the names of the flags — in both cases, the default value is the zero value for the associated type. Is your claim that the name of the flag is meaningful by itself? If so, I reject the claim 🙃 and request that you remove this helper in favor of Pair. If there is another technical reason, please correct my understanding!

fftest/vars.go
aramcodz: `PresetValsFS() ` is no longer needed and has been removed.

ffyaml/ffyaml_test.go
aramcodz: Test has been removed. see latest.

ffyaml/ffyaml.go
aramcodz: UPDATE.

@peterbourgon, As per your suggestion/argument in favor of blank strings, I modified the approach of my suggested YAML parser fix/change. It now will always return an empty string ("") for the "nil-case". 

Important: the implication is that `Parse()` will now return Parse errors if the blank YAML entry is for a config field that is of a type other than string or slice (of strings). See new/updated unit test function `TestParserReturnsErrorsForBlankNonStrings()`

Please see latest commits:
1. https://github.com/aramcodz/ff/commit/75e3485c94788728dddff61ab48e50436849294e
2. just fixed imports in ffyaml_test.go -> https://github.com/aramcodz/ff/commit/5e9383ff9ba0f3f4f0f09b5cc9d4fcd12aa8a14d

Is this solution to the YAML parser issue with blank fields acceptable?

ffyaml/ffyaml_test.go
aramcodz: Sorry. I have added in an auto-run `goimports` entry now upon save.

ffyaml/ffyaml.go
peterbourgon: > The implication is that Parse() will now return Parse errors if the blank YAML entry is for a config field

Rather, a flag name.

> that is of a type other than string or slice (of strings). 

OK. Should it? 

Every flag value starts as a string and is converted to the flag's type. The JSON parser stringifies all of its values before passing them on. This will happen to the YAML values, too! What you're doing here is an extra layer of validation, which is then repeated, essentially, by the flag parsing code.

Does that all make sense? Do you still want to validate types at the parser level?

aramcodz: > What you're doing now is an extra layer of type validation at the file parser level, and before the flag parsing code runs. The JSON parser, for example, doesn't do this — it stringifies all of its values before passing them on, so doesn't do this kind of thing.

@peterbourgon: I am really confused by your response. I did not add in **_any** type validation_ into the YAML parser.  Literally, to the actual implementation (non-test) ff code, the only change I am suggesting is adding in to `ValToStr()` function, lines 73-74:

```
	case nil:
		return "", nil

```

That 2-line addition is now the only proposed change in this PR to any implementation code in ff. 

And, in the test code, the most significant change I made is adding in new test function `TestParserReturnsErrorsForBlankNonStrings()`. This new test function demonstrates that it is indeed the `Parse()` function in `parse.go`, Not anything in `ffyaml.go` `Parser()`, that returns parse errors for blank YAML entries if the pre-defined Flagset type is something other than a string or slice of strings.  

So, I am really confused. 

Specifically, what is it in the implementation code you actually want me to change to fix the blank YAML entry error problem? 

What is it you will accept in a PR as a change to fix this issue with blank entries the YAML parser?

peterbourgon: Sorry, I should have clarified. I understand that the current ffyaml parser already does this validation, and that you haven't changed anything in that regard. I'm asking if it should _continue_ doing that validation. If your position is that your change doesn't affect the current behavior and so you don't care, then that's fine 👍

aramcodz: >  If your position is that your change doesn't affect the current behavior and so you don't care, then that's fine.

I am focused on fixing the blank entries in YAML files issue. That's it. That is my priority because my team at IBM Watson Health has been developing some software and has been using **ff** as a dependency in our most recent update one of our services, specifically an overhaul to this (older, Preview Version) https://github.com/Alvearie/hri-mgmt-api. 

My PR really all about solving that problem (or, perhaps, at least changing the current behavior to be much more user friendly) with a small, incremental, least intrusive, change to the ff YAML parser. 

Making that one change in ff means it meets all our needs as a config library. That is, currently, this behavior (with the blank YAML entries) causes an issue for us during deployments that our users cannot understand exactly why the deployment failed. 

Hence, really all I am trying to accomplish here is a newly released version of ff, with the Blank-Entries-YAML-Files issue behavior "fixed", behaving in a way seems like a significant improvement for end-users over current behavior. 

We would _prefer_ contributing this proposed fix/improvement back the the **ff** OS project, as good Open Source community members, rather than just writing our own version of the parser itself.

peterbourgon: All good. Sorry for the distraction, then.

ffyaml/ffyaml.go
```diff
@@ -68,6 +71,8 @@ func valToStr(val interface{}) (string, error) {
 		return strconv.FormatInt(v, 10), nil
 	case float64:
 		return strconv.FormatFloat(v, 'g', -1, 64), nil
+	case nil:
+		return "", nil
```
peterbourgon, line 75: So now this is the only substantive change in the PR?

aramcodz: That is correct. Everything else is just Unit test or related to try to self-document what the behavior is in the case of Blank YAM entries to FlagSet entries of various types. 

I am open to amending the unit test & fixture code. But, my whole goal with that stuff was to be able show to other lib users of ff, what the expected YAML parser behavior is with this particular test case, i.e, classic T-D-D, unit test/fixture code as clear, explicit documentation of one particular behavioral edge-case...that I happened to run into.

PR #84, opened 2021-07-11
Set up Dependabot

mmorel-35:


peterbourgon:
Dependencies should not be updated automatically.

mmorel-35:
They aren't, this creates a PR for each dependency update that you can always choose to accept or reject.

peterbourgon:
Don't want that either.

mmorel-35:
Fine.

PR #94, opened 2022-05-20
Update yaml.v2 and go-toml modules

decke:
Generated with:
```
go get -u ./...
go mod tidy
go test
```

Builds fine and tests run fine as well. Runtime testing was limited to my own use case.

This also brings yaml.v2 to a version which is not affected by [CVE-2019-11254](https://nvd.nist.gov/vuln/detail/CVE-2019-11254) anymore. For details also see: https://github.com/advisories/GHSA-wxc4-f4m6-wwqv

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,6 @@ module github.com/peterbourgon/ff/v3
 go 1.16
 
 require (
-	github.com/pelletier/go-toml v1.6.0
-	gopkg.in/yaml.v2 v2.2.4
+	github.com/pelletier/go-toml v1.9.5
+	gopkg.in/yaml.v2 v2.4.0
 )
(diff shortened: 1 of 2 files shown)


peterbourgon:
So, if I [understand correctly](https://github.com/go-yaml/yaml/commit/53403b58ad1b561927d19068c655246f2db79d48), the issue was excessive resource consumption from pathological .yaml input?

decke:
For reference I also tried to update ff to yaml.v3 3.0.1 and it works fine as well. Please let me know if you prefer that and I can update the PR if you want.

decke:
I know all of this but it is not an effort to do automatic updates. The reasons behind it are to avoid known vulnerabilities.

peterbourgon:
> I know all of this but it is not an effort to do automatic updates. The reasons behind it are to avoid known vulnerabilities.

My understanding was that your linked CVEs don't actually apply to the yaml and toml dependencies used by this module?

I mean, no big deal in this case.

decke:
Thanks for merging.

PR #95, opened 2022-05-26
Fix example demonstrating usage of env vars

m90:
It seems `WithEnvVars` does not exist in version 3, but instead `WithEnvVarNoPrefix` is now doing what is described in the example.


peterbourgon:
Oops! Thanks!

Issue #105, opened 2023-05-03
ffcli.DefaultUsageFunc: Match output of FlagSet.PrintDefaults


abhinav:
This issue is to discuss a possible change in behavior in advance of a PR.
Currently, `-h` on a `ffcli.Command` prints a FLAGS section like the following:

```
  -b=false  bool
  -d 0s     time.Duration
  -f 0      float64
  -i 0      int
  -s ...    string
  -x ...    collection of strings (repeatable)
```

Things to note:
- Flags use the placeholder "..." instead of "string", "int", or something more helpful.
- Flags are always on the same line as the flag usage. Usage text has to shift right if a flag with a different length is added.
- Defaults are printed in the help message instead of being put at the end of the usage text, which adds another case when usage text has to shift right.
- Usage text shifting right adds uncertainty to how much horizontal space is available to a help message for a flag.
- Doesn't handle newlines in the flag usage text; it'll wrap to the next line instead of being indented to match.

FlagSet.PrintDefaults, on the other hand, prints:

```
  -b        bool
  -d duration
            time.Duration
  -f float
            float64
  -i int
            int
  -s string
            string
  -x value
            collection of strings (repeatable)
```

Only short boolean flags are on the same line -- if they fit in that space. All other usage text is on the next line and indented.

Not visible in this example:
- non-zero defaults are added at the end in the form `(default: foo)`
- newlines in usage text are adjusted to match indentation of the rest of the text
- if the help text wraps a word with backticks, for example:

    ```
    path to input `file`
    ```

  then that word is treated as the placeholder in the help:

    ```
    -f file
              path to input file
    ```

It's not terribly difficult to change DefaultUsageFunc to match this behavior, but it is a change in behavior.
Are you open to such a change or would you prefer that users write a custom UsageFunc for this?

peterbourgon:
I'm happy to make changes like the one that was merged, but I definitely don't want ff.DefaultUsageFunc to match flag.FlagSet.PrintDefaults — I consider the latter to be pretty user-unfriendly, overall.

peterbourgon:
> Split flag and help across two lines if flag is long — you've indicated this is undesirable

👍 

> Support multi-line flag help message — I suspect this would be acceptable

I'd say this is out of scope for DefaultUsageFunc, and something you should solve with your own implementation. (IMO multi-line flag help messages are a code smell, so to speak, and should be fixed, though I understand reasonable people may disagree.)

[...]

PR #106, opened 2023-05-03
ffcli.DefaultUsageFunc: Support flag help placeholders

abhinav:
(This is an arguably uncontroversial part of the proposal in #105.
Its's a new feature, not a significant change in behavior.)

Currently, ffcli.DefaultUsageFunc prints "..." for any flag
that does not have a default value specified.
This produces less-than-effective help from DefaultUsageFunc.

This change retains the behavior of printing the default value as-is,
but if a default value is not provided,
it allows users to provide placeholder text
by wrapping a word inside the help text for a flag in backticks.

For example, given the following:

    fset.String("c", "" /* default */, "path to `config` file")

We'll get:

    -c config  path to config file

This matches the behavior of FlagSet.PrintDefaults,
and indeed it relies on the same flag.UnquoteUsage machinery for this.

This also has the nice side-effect of making a reasonable guess
at an alternative placeholder text instead of "...". For example:

    fset.Int("n", "" /* default */, "number of items")

    // Before: -n ...  number of items
    // Now:    -n int  number of items

Note that as implemented right now, the user supplied placeholder will
be used only if a non-zero default value was not supplied.
This was an attempt to retain as much of the existing behavior.
The proposal in #105, if you're open to it,
would change more of the output.

### DIFF
--- a/ffcli/command.go
+++ b/ffcli/command.go
@@ -231,12 +231,33 @@ func DefaultUsageFunc(c *Command) string {
 				space = "="
 			}
 
-			def := f.DefValue
-			if def == "" {
+			// If the help text contains backticks,
+			// e.g. "foo `bar` baz"`, we'll get:
+			//
+			//   argname = "bar"
+			//   usage   = "foo bar baz"
+			//
+			// Otherwise, it's an educated guess for a placeholder,
+			// or an empty string if one couldn't be determined.
+			argname, usage := flag.UnquoteUsage(f)
+
+			// For the argument name printed in the help,
+			// the order of preference is:
+			//
+			//  1. the default value
+			//  2. the back-quoted name from the help text
+			//  3. the '...' placeholder
+			var def string
+			switch {
+			case f.DefValue != "":
+				def = f.DefValue
+			case argname != "":
+				def = argname
+			default:
 				def = "..."
 			}
 
-			fmt.Fprintf(tw, "  -%s%s%s\t%s\n", f.Name, space, def, f.Usage)
+			fmt.Fprintf(tw, "  -%s%s%s\t%s\n", f.Name, space, def, usage)
 		})
 		tw.Flush()
 		fmt.Fprintf(&b, "\n")
--- a/ffcli/command_test.go
+++ b/ffcli/command_test.go
@@ -439,6 +439,62 @@ func TestIssue57(t *testing.T) {
 	}
 }
 
+func TestDefaultUsageFuncFlagHelp(t *testing.T) {
+	t.Parallel()
+
+	for _, testcase := range []struct {
+		name string // name of test case
+		def  string // default value, if any
+		help string // help text for flag
+		want string // expected usage text
+	}{
+		{
+			name: "plain text",
+			help: "does stuff",
+			want: "-x string  does stuff",
+		},
+		{
+			name: "placeholder",
+			help: "reads from `file` instead of stdout",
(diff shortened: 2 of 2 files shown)


peterbourgon:
I'm open to allowing users to specify placeholder text other than `...`, but I'm not convinced that text in backticks in the help string is a reasonable way to encode that information.

ffcli/command.go
```diff
@@ -231,12 +231,25 @@ func DefaultUsageFunc(c *Command) string {
...
 
+			// If the help text contains backticks,
+			// e.g. "foo `bar` baz"`, we'll get:
+			//
+			//   argname = "bar"
+			//   usage   = "foo bar baz"
+			//
+			// Otherwise, it's an educated guess for a placeholder,
+			// or an empty string if one couldn't be determined.
+			argname, usage := flag.UnquoteUsage(f)
```
peterbourgon, line 242: This appears to be the major change, is there some prior art that uses `UnquoteUsage` in this way?

abhinav:
RE: Prior art: this is [how the std flag package does it](https://cs.opensource.google/go/go/+/refs/tags/go1.20.4:src/flag/flag.go;l=575-580). This is documented mostly in [flag.PrintDefaults](https://pkg.go.dev/flag#PrintDefaults):

> ```
> -x int
> 	usage-message-for-x (default 7)
> ```
> 
> [..] The listed type, here int, can be changed by placing a back-quoted name in the flag's usage string; the first such item in the message is taken to be a parameter name to show in the message and the back quotes are stripped from the message when displayed. For instance, given
>
> ```
> flag.String("I", "", "search `directory` for include files")
> ```
> 
> the output will be
>
> ```
> -I directory
> 	search directory for include files.
> ```

This PR isn't suggesting a completely custom behavior; just to carry over something that the std flag package already provides. I'll agree that this feature isn't super obviously documented in std flag, but users of flag who know about it will likely expect it from ffcli as well. (That's what brought me here! 😄 )

About whether this is the best way to encode this information:
I have mixed feelings about it. While I like the idea of incorporating the variable name in the help message when possible, I would have preferred for std flag to have a more explicit means to set this when it's not. For example, by setting a field on flag.Flag, or implementing a method on a flag.Value. That might be worth pursuing with a proposal upstream, but I don't have the appetite for that right now.

So right now, this is just suggesting matching the behavior of what flag.PrintDefaults already provides.

peterbourgon:
Oh, cool! Somehow I didn't know about that.

ffcli/command.go
```diff
@@ -231,12 +231,25 @@ func DefaultUsageFunc(c *Command) string {
...
 			def := f.DefValue
 			if def == "" {
-				def = "..."
+				if argname != "" {
+					def = argname
+				} else {
+					def = "..."
+				}
```
peterbourgon, line 249:
```go
var def string
switch {
case f.DefValue != "":
	def = f.DefValue
case f.DefValue == "" && argname != "":
	def = argname
default:
	def = "..."
}
```

?

abhinav: Reasonable

abhinav:
@peterbourgon Obviously no rush, but FYI, the suggested change has been made.

ffcli/command.go
peterbourgon: Apparently flag.FlagSet lol 😮‍💨

peterbourgon:
Thanks for the poke.

Issue #107, opened 2023-05-23
Nested Config


lmittmann:
Would you be open to a PR that enables nesting inside JSON configurations? This would allow flags containing dots to be looked up at the corresponding path. E.g. the flag `a.b` could be configured inside `{"a": {"b":"xxx"}}`.

peterbourgon:
Probably not.

lmittmann:
This is possible with `fftoml`, why not allow it for json and yml?

peterbourgon:
I think it may have been a mistake to allow it in `fftoml`.

But, upon reflection, I suspect my initial reaction may be too conservative, and there may be actual value in this feature.

I'm open to reviewing a change that allows nesting of configuration in this way, provided the API and behavior is consistent across ffyaml, fftoml, and ffjson.

jolheiser:
This can be closed per #112 

Thanks again to both of you!

PR #112, opened 2023-07-18
JSON, TOML, YAML: support for nested config data

peterbourgon:
This PR is a simple adaptation of the TraverseMap function from https://github.com/jolheiser/ff/pull/1 (thanks, @jolheiser!) which is applied to the JSON, TOML, and YAML config file parsers more or less in the same way. It allows nested config data in any of those formats to map to flag names in a reasonably well-defined way.

Addresses #107 and #108. I'm pretty sure this isn't a breaking change, but I'd appreciate a review from anyone who'd care to give it a look.

This is a precursor to a much larger refactor, including many highly requested features, coming soon to theaters near you.

### DIFF
--- a/fftoml/fftoml.go
+++ b/fftoml/fftoml.go
@@ -2,144 +2,36 @@
 package fftoml
 
 import (
-	"fmt"
 	"io"
-	"strconv"
 
 	"github.com/pelletier/go-toml"
-	"github.com/peterbourgon/ff/v3"
+	"github.com/peterbourgon/ff/v3/internal"
 )
 
-// Parser is a parser for TOML file format. Flags and their values are read
-// from the key/value pairs defined in the config file.
+// Parser is a helper function that uses a default ParseConfig.
 func Parser(r io.Reader, set func(name, value string) error) error {
-	return New().Parse(r, set)
+	return (&ParseConfig{}).Parse(r, set)
 }
 
-// ConfigFileParser is a parser for the TOML file format. Flags and their values
-// are read from the key/value pairs defined in the config file.
-// Nested tables and keys are concatenated with a delimiter to derive the
-// relevant flag name.
-type ConfigFileParser struct {
-	delimiter string
+// ParseConfig collects parameters for the TOML config file parser.
+type ParseConfig struct {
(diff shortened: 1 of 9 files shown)


internal/traverse_map.go
```diff
@@ -0,0 +1,53 @@
...
+		return set(key, strconv.FormatBool(v))
+	case nil:
+		return set(key, "")
+	case []any:
+		for _, v := range v {
+			if err := traverseMap(key, v, delimiter, set); err != nil {
+				return err
+			}
+		}
+	case map[string]any:
```
jolheiser, line 40: This will also require the change I made https://github.com/peterbourgon/ff/pull/108/commits/47743d74e9cc31bc9210a2b119db224ab20b83e0 to handle `map[any]any` which is what the YAML decoder uses as the default base type.

peterbourgon: Interesting, can you suggest a test case that would have caught this?

jolheiser: A similar test as the one added in #108 
```yaml
foo:
  s: a string
  b: true
bar:
  nested:
    i: 123
    f: 1.23
baz:
  nested:
    x:
      - one
      - two
      - three
 ```

File: fftoml/fftoml.go [hunk 2]
```diff
@@ -62,3 +67,19 @@ func WithTableDelimiter(d string) Option {
 		c.delimiter = d
 	}
 }
+
+// ParseError wraps all errors originating from the TOML parser.
+type ParseError struct {
```
lmittmann, line 72: If you decide to introduce a breaking change at some point I would suggest dropping the `type ParseError struct` for something like:

```go
var ParseError = errors.New("error parsing TOML config")

// ...

func (c ConfigFileParser) Parse(...) {
    // ...
    if err := toml.NewDecoder(r).Decode(&m); err != nil {
        return fmt.Errorf("%w: %w", ParseError, err)
    }
    // ...
```
You would also need to upgrade to go1.20 for this change.

peterbourgon: It was a mistake to introduce error types to this package. The next version will absolutely remove them.

Issue #72, opened 2020-11-28
ffcli: Default usage messages of subcommands lack information from preceding commands in the tree


ChrisHines:
Run the program below (or on the [playground](https://play.golang.org/p/_e_VqwPo-s7)) to see an example of the problem. 

```go
package main

import (
	"flag"

	"github.com/peterbourgon/ff/v3/ffcli"
)

func main() {
	barfs := flag.NewFlagSet("bar", flag.ContinueOnError)
	barfs.String("bf", "", "bar flag")
	bar := &ffcli.Command{
		Name:      "bar",
		FlagSet:   barfs,
		ShortHelp: "bar help",
	}

	foofs := flag.NewFlagSet("foo", flag.ContinueOnError)
	foofs.String("ff", "", "foo flag")
	foo := &ffcli.Command{
		Name:        "foo",
		FlagSet:     foofs,
		ShortHelp:   "foo help",
		Subcommands: []*ffcli.Command{bar},
	}

	rootfs := flag.NewFlagSet("root", flag.ContinueOnError)
	rootfs.String("rf", "", "root flag")
	root := &ffcli.Command{
		Name:        "root",
		FlagSet:     rootfs,
		ShortHelp:   "root help",
		Subcommands: []*ffcli.Command{foo},
	}

	root.Parse([]string{"foo", "bar", "-h"})
}
```

It prints:

```
USAGE
  bar

FLAGS
  -bf ...  bar flag
```

It should be obvious that the above help information is inadequate. The bar command is presented detached from its context as a subcommand of `foo` which is likewise a subcommand of `root`.

I suggest that the help information should look more like this:

```
USAGE
  root [flags] foo [flags] bar [flags]

root FLAGS
  -rf ...  root flag

foo FLAGS
  -ff ...  foo flag

bar FLAGS
  -bf ...  bar flag
```

We can write a custom `UsageFunc` to produce that help text, but that doesn't scale, is tedious to maintain, and couples subcommand metadata tightly to its position in the command tree. I suggest that `ffcli` should be able to create the above help text from only the metadata provided in the program shown above.

I have prototyped a change that achieves that goal and will submit a PR showing that work for discussion. Note: The change requires a breaking API change.

ChrisHines:
My prototype is in #73. I have submitted it as a Draft since I have made no effort to update any documentation to reflect the API changes.

peterbourgon:
[...]

PR #113, opened 2023-07-20
v4.0.0-alpha.1

peterbourgon:
This PR introduces v4 of package ff, which includes several enhancements and breaking changes.

## What are the major changes?

- A new [Flags](https://github.com/peterbourgon/ff/blob/v4-alpha.1/flags.go) interface is defined to represent a flag set
- A new [Flag](https://github.com/peterbourgon/ff/blob/v4-alpha.1/flags.go) interface is defined to represent a single flag
- A getopts-style [CoreFlags](https://github.com/peterbourgon/ff/blob/v4-alpha.1/flags_core.go) implementation of the Flags interface is provided
- (You can still pass a stdlib flag.FlagSet to Parse, in which case it's converted to a CoreFlags)
- CoreFlags supports the concept of parent flag sets, making it easier to build hierarchical command trees
- [package ffval](https://github.com/peterbourgon/ff/blob/v4-alpha.1/ffval) provides a set of common flag values, and helpers for constructing your own flag values
- ffcli is merged into ff
- [Command](https://github.com/peterbourgon/ff/blob/v4-alpha.1/command.go) has fewer fields
- Commands can optionally be Reset, allowing the same command tree to be parsed more than once
- Usage is refactored to be a free function, taking a command or flag set, and returning a string
- [package ffhelp](https://github.com/peterbourgon/ff/blob/v4-alpha.1/ffhelp) provides default usage funcs, and helpers for writing your own usage funcs
- The JSON config file parser is moved to [package ffjson](https://github.com/peterbourgon/ff/blob/v4-alpha.1/ffjson)
- The .env file parser is moved to [package ffenv](https://github.com/peterbourgon/ff/blob/v4-alpha.1/ffenv)
- fftoml is updated to use [pellegier/go-toml@v2](https://pkg.go.dev/github.com/pelletier/go-toml/v2)

## Why move away from flag.FlagSet?

Basically, it was too limited.

While I'm not necessarily sold on all of getopts(3), I think allowing flags to have long (--foo) and short (-f) names is an important feature, and a basic expectation for nearly all users. The fact that the stdlib flag set doesn't support this feature proved to be a problem worth fixing.

Also, using a concrete type kind of paints the module into a corner, in terms of capability and extensibility. I don't think it would have been possible to provide the most-requested features (hierarchical flags, support for different flag set implementations, etc.) without switching to some kind of abstraction.

## Why adapt a flag.FlagSet to a bespoke CoreFlags instead of using a separate type?

To take advantage of CoreFlags features like Reset and SetParent, without needing to re-implement them. (I'm not totally convinced this is the right approach, but it's where I've landed for the moment.)

## Why model hierarchy as parent flag sets instead of multiple flag sets in a command?

You always have a single set of args provided by the user. If you want to apply those args to multiple flag sets, then you need to find a way for the Parse method of each flag set to consume only the args that map to flags known to the flag set, and yield the args that map to flags that aren't known to the flag set, while also returning non-flag args separately. The command parser would then need to call Parse for each flag set in sequence, passing down the leftover args after each stage. 

This would require changes to the Parse method in the Flags interface, which is currently extremely good and intuitive: `Parse(args []string) error`. Complicating it would be a very tall order, and a large burden for implementations. And after a lot of experimentation, I concluded that it was infeasible to reliably distinguish "a flag that we don't know about" from "a non-flag argument" when parsing args.

Consequently, the design keeps the simpler Parse signature in the Flags interface, and moves all of the logic related to parent flags into the implementation(s). The GetFlags method in the Flag interface was added in part so that callers can know in which flag set a flag was defined.


examples/objectctl/pkg/createcmd/create.go
```diff
@@ -0,0 +1,55 @@
...
+	*rootcmd.RootConfig
+	Overwrite bool
+	Flags     *ff.CoreFlags
+	Command   *ff.Command
+}
+
+func New(rootConfig *rootcmd.RootConfig) *CreateConfig {
+	var cfg CreateConfig
+	cfg.RootConfig = rootConfig
+	cfg.Flags = ff.NewFlags("create").SetParent(cfg.RootConfig.Flags)
```
peterbourgon, line 23: I don't want to make SetParent part of the Flags interface, because I don't want to dictate that flag sets need to support the concept of parents. At the moment it's a method on the CoreFlags type, and takes a parent CoreFlags value as a parameter. That means that you need to keep track of the concrete CoreFlags value of parent commands somehow, you can't just rely on the Flags field in the parent command directly. At least, not without type assertions. This is annoying, but I'm not sure what a better approach would be.

parse.go
```diff
@@ -11,311 +11,301 @@ import (
...
-	var c Context
-	for _, option := range options {
-		option(&c)
+// Parse the flag set with the provided args. [Option] values can be used to
+// influence parse behavior. For example, options exist to read flags from
+// environment variables, config files, etc.
+//
+// The fs parameter must be of type [Flags] or [*flag.FlagSet]. Any other type
+// will result in an error.
+func Parse(fs any, args []string, options ...Option) error {
```
peterbourgon, line 20: I'm not super happy with `fs any` but I judge that maintaining support for `ff.Parse(<*flag.FlagSet>, ...)` is important enough to justify complecting the API in this way. I could be convinced otherwise, especially if there were a better design that I haven't thought of...

command.go
```diff
@@ -0,0 +1,234 @@
...
+	// Flags is the set of flags associated with, and parsed by, this command.
+	//
+	// When building a command tree, it's often useful to allow flags defined by
+	// parent commands to be specified by any subcommand. The core flag set
+	// supports this behavior via SetParent, see the documentation of that
+	// method for details.
+	//
+	// Optional. If not provided, an empty flag set will be constructed and used
+	// so that the -h, --help flag works as expected.
+	Flags Flags
```
peterbourgon, line 62: Command.Flags needs to be a true Flags, meaning you can't pass a flag.FlagSet directly, you have to use the adapter constructor. I think this is reasonable: ff.Parse is meant to be symmetrical with flag.Parse, but ff.Command doesn't have any direct analogue in that sense.

command.go
```diff
@@ -0,0 +1,234 @@
...
+
+	return cmd.selected.GetSelected()
+}
+
+// GetParent returns the parent command of this command, or nil if a parent
+// hasn't been set. Parents are set during the parse phase, but only for
+// commands which are traversed.
+func (cmd *Command) GetParent() *Command {
+	return cmd.parent
+}
```
peterbourgon, line 185: The intent here is to help usage functions that want to include [information from preceding commands in the tree](https://github.com/peterbourgon/ff/issues/72).

parse.go
peterbourgon: I would have loved to use generics to say `fs interface{ ff.FlagSet | *flag.FlagSet }` — but you can't define a union over anything but concrete types. I guess it makes sense why, but it's unfortunate in this case.

mfridman:
More of a drive-by. Are you looking for feedback on this `/v4`?

I've used the `/v3` fairly extensively in a few projects, so looking forward to whatever iteration comes next.

peterbourgon:
Feedback is very much appreciated!

peterbourgon:
I'm going to go ahead and release this as `v4.0.0-alpha.1` so it can be imported. Feedback still appreciated.

parse.go
flowchartsman: I agree that seeing `any` in a signature where there is actually a runtime constraint is uncomfortable, but that's at least partially due to the permissive connotation. While we can't (_yet_) be more specific with general interfaces, you might at least be able to sharpen up the perceptual edges with a type alias like:

```
// AnyFlagSet must be one of { [ff.FlagSet] | [*flag.FlagSet] }
type AnyFlagSet = any
```

This at least calls out the restriction for now, even if it doesn't have any compile-time checking.

flowchartsman: Or, alternatively, `FlagSetAny`, depending on whether you think of it as "an `any` that's a flagset" or "[the] FlagSet types' `any`" :D

peterbourgon: I like this a lot!

flowchartsman: #115 for your perusal!

PR #115, opened 2023-09-04
Type alias for runtime-constrained flag set `any`

flowchartsman:
Having an `any` in the signature for `Parse` is unsatisfying because it
is actually a runtime-constrained value, and `any` does not convey this
information.

In lieu of support for "union type" general interfaces which allow the
intermixing of basic interfaces with concrete types, a type alias to
`any` with a descriptive name at least places the constraint
front-and-center.

### DIFF
--- a/parse.go
+++ b/parse.go
@@ -11,13 +11,16 @@ import (
 	"strings"
 )
 
+// FlagSetAny must be one of { [Flags] | [*flag.FlagSet] }
+type FlagSetAny = any
+
 // Parse the flag set with the provided args. [Option] values can be used to
 // influence parse behavior. For example, options exist to read flags from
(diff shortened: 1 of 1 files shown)


flowchartsman:
Aha, never mind! My clever ploy to get on your committer's list with a four-line PR is thwarted! ;) (ref: b12d8988967b158339daad0d82a1c0ba4ba63f0f)

peterbourgon:
> Aha, never mind! My clever ploy to get on your committer's list with a four-line PR is thwarted! ;) (ref: [b12d898](https://github.com/peterbourgon/ff/commit/b12d8988967b158339daad0d82a1c0ba4ba63f0f))

😭

PR #119, opened 2023-09-18
fix: WithFilesystem option should take a io.FS as a parameter

ralf-cestusio:
Taking the FS interface instead of the concrete implementation of embed.FS allows passing any kind of filesystem


ralf-cestusio:
Closes #121

PR #129, opened 2024-01-27
Allow env var split delimiter escaping

### LINKED ISSUE #127: Env var split with escape character ?
hi! thanks for ff, i've been using it for [gonic](https://github.com/sentriz/gonic) for years 

there was a user today with an env var such as `GONIC_MPV_ARGS="--audio-device=alsa/iec958:CARD=ta,DEV=0"`

which resulting in an unwanted split. i know it can turned off or change the delimiter, but gonic does use the comma splitting for other flags already

i wonder is it feasible to comma escape the demilimiter like `a\,b`? 

thanks!

peterbourgon:
Should close #127.

### DIFF
--- a/options.go
+++ b/options.go
@@ -114,7 +114,7 @@ func WithEnvVarPrefix(prefix string) Option {
 // one time, with the value `a,b,c`. Providing WithEnvVarSplit with a comma
 // delimiter would set `foo` multiple times, with the values `a`, `b`, and `c`.
 //
-// If an env var value contains the delimeter prefixed by a single backslash,
+// If an env var value contains the delimiter prefixed by a single backslash,
 // that occurrence will be treated as a literal string, and not as a split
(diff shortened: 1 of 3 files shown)


options.go
```diff
@@ -114,6 +114,12 @@ func WithEnvVarPrefix(prefix string) Option {
 // one time, with the value `a,b,c`. Providing WithEnvVarSplit with a comma
 // delimiter would set `foo` multiple times, with the values `a`, `b`, and `c`.
 //
+// If an env var value contains the delimeter prefixed by a single backslash,
```
peterbourgon, line 117:
```suggestion
// If an env var value contains the delimiter prefixed by a single backslash,
```

peterbourgon:
I'm not sure if this behavior should be gated behind an option. I can't immediately think of a (reasonable) situation where this change would break anyone...

PR #134, opened 2024-04-09
add {Type}LongVar and {Type}ShortVar helpers

lopezator:
Add helper functions to avoid repeating the `0` and/or `""` in the case of `Var` flagset method calls.

### DIFF
--- a/flag_set.go
+++ b/flag_set.go
@@ -776,6 +776,18 @@ func (fs *FlagSet) BoolVar(pointer *bool, short rune, long string, usage string)
 	return fs.BoolVarDefault(pointer, short, long, false, usage)
 }
 
+// BoolShortVar defines a new bool flag in the flag set, and panics on any
+// error.
+func (fs *FlagSet) BoolShortVar(pointer *bool, short rune, def bool, usage string) Flag {
+	return fs.Value(short, "", ffval.NewValueDefault(pointer, def), usage)
+}
+
+// BoolLongVar defines a new bool flag in the flag set, and panics on any
+// error.
+func (fs *FlagSet) BoolLongVar(pointer *bool, long string, def bool, usage string) Flag {
+	return fs.Value(0, long, ffval.NewValueDefault(pointer, def), usage)
+}
+
 // BoolVarDefault defines a new bool flag in the flag set, and panics on any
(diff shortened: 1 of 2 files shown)


lopezator:
I tend to use just long flag names for all my applications, so moving out from `flag.FlagSet` to `ff.FlagSet` means in my case having to add a `0` to every line, like:

```go
fs.StringVar(&cfg.GRPCAddr, 0, "grpc-addr", "localhost:8000", "gRPC address")
fs.StringVar(&cfg.HTTPAddr, 0, "http-addr", "localhost:8001", "HTTP address")
// And so on...
``

As `StringShort` and `StringLong` were already present, I took the opportunity to add the short method as well, so others can benefit from it. It follows the same approach as the mentioned ones, just porting the functionality to the `Var` methods.

I suppose you didn't add this before to avoid this implementation for every type, but in my case at least, string is like the 90% of the the flags. 

If you feel that we should add support for all types, I can do that as well. Just wanted to start small and simple.

Thanks for the wonderful library btw @peterbourgon

peterbourgon:
So I'm not totally convinced that the benefit from expanding the API surface area to include these new methods outweighs the cost of adding these zero-value params, but I guess it could make sense in a holistic way. Can you add the same variants to the other type-specific XxxVar methods on FlagSet?

lopezator:
> So I'm not totally convinced that the benefit from expanding the API surface area to include these new methods outweighs the cost of adding these zero-value params

To be honest, I felt the same way, but as the concession is already made on the flagSet types without var (`fs.String`, `fs.Bool`...) I thought it might make sense adding it to the `xxxVar` variants as well fo the sake of consistency.

Another option could be to remove all the `Short` and `Long` variants everywhere? But I guess that would require more work and also breaks the API, but it's maybe acceptable since this is not a stable release yet, I don't know :slightly_smiling_face: 

I added the rest of the types along with two new tests.

Thanks a lot for the quick review and let me know if I can help in another way.

lopezator:
No hurries, obviously, but if there is there anything else I can improve on my end, just tell me @peterbourgon .

Thanks!

peterbourgon:
Some changes related to this PR have been merged in the interim.

PR #136, opened 2024-06-14
Support default for flag.Value implementations registered via AddStruct

GeorgeMac:
I noticed when I had a field which implemented `flag.Value` and I registered it using `AddStruct` that the default wasn't being respected. This adds support this by wrapping the `flag.Value` in another implementation which tracks if `Set` is called and returns the default if not.


peterbourgon:
PTAL at #143

PR #137, opened 2024-11-27
Fix ErrDuplicateFlags when no env var parsing is requested

tgulacsi:
Same cased short flags should be possible, but ErrDuplicateFlags is returned even when environment variable parsing is not requested.
Do not check for duplicate folds if no env var is requested.

### DIFF
--- a/parse_test.go
+++ b/parse_test.go
@@ -449,10 +449,10 @@ func TestParse_shortSameCase(t *testing.T) {
 	})
 	t.Run("WithEnv", func(t *testing.T) {
 		t.Parallel()
-		if err := ff.Parse(newFS(), args, ff.WithEnvVars()); err == nil {
-			t.Error("wanted ErrDuplicateFlage, got nil")
-		} else if !errors.Is(err, ff.ErrDuplicateFlag) {
-			t.Errorf("wanted ErrDuplicateFlage, got %+v", err)
+		if err := ff.Parse(
+			newFS(), args, ff.WithEnvVars(),
+		); err == nil || !errors.Is(err, ff.ErrDuplicateFlag) {
+			t.Errorf("wanted ErrDuplicateFlag, got %+v", err)
 		}
 	})
 }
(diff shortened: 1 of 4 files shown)


parse_test.go
```diff
@@ -428,3 +429,30 @@ func TestParse_stdfs(t *testing.T) {
...
+		if err := ff.Parse(newFS(), args); err != nil {
+			t.Error(err)
+		}
+	})
+	t.Run("WithEnv", func(t *testing.T) {
+		t.Parallel()
+		if err := ff.Parse(newFS(), args, ff.WithEnvVars()); err == nil {
+			t.Error("wanted ErrDuplicateFlage, got nil")
+		} else if !errors.Is(err, ff.ErrDuplicateFlag) {
+			t.Errorf("wanted ErrDuplicateFlage, got %+v", err)
```
ccoVeille, line 455: Do you need the first if?


```suggestion
		if err := ff.Parse(newFS(), args, ff.WithEnvVars()); !errors.Is(err, ff.ErrDuplicateFlag) {
			t.Errorf("wanted ErrDuplicateFlag, got %+v", err)
```

I mean the %+v will report something for nil: `<nil>`

tgulacsi: Thanks

parse_test.go
```diff
@@ -428,3 +429,30 @@ func TestParse_stdfs(t *testing.T) {
...
+		t.Parallel()
+		if err := ff.Parse(newFS(), args); err != nil {
+			t.Error(err)
+		}
+	})
+	t.Run("WithEnv", func(t *testing.T) {
+		t.Parallel()
+		if err := ff.Parse(
+			newFS(), args, ff.WithEnvVars(),
+		); err == nil || !errors.Is(err, ff.ErrDuplicateFlag) {
```
ccoVeille, line 454: Isn't this enough?

```suggestion
		); !errors.Is(err, ff.ErrDuplicateFlag) {
```

I'm just curious, I'm fine with the test you have, it's clearer

ccoVeille (approved):

peterbourgon:
I believe this should be fixed by #153 -- please re-open if that's not the case.

PR #139, opened 2025-01-17
fix: --long= should not consume the next argument

telemachus:


telemachus:
I added a couple of tests to confirm that (1) for a boolean, `--long=` parses as true, (2) for other flags `--long=` parses as `--long=""`, and (3) `--long=` does not consume the next argument from `args`.

In the parsing code, I tried to minimize changes, but while I was there I updated two comments to reflect that only long flags were at issue, and I changed `len(args) <= 0` to `len(args) == 0` since I don't think it's ever possible for a slice to have a negative length.

I considered adding a further test for `--long=""` with non-string flags, but I wasn't sure whether testing for such a specific error was too fiddly. Let me know if you think more tests are needed.

```go
{
	Name:         "--flt= -a",
	Constructors: []fftest.Constructor{fftest.CoreConstructor},
	Args:         []string{`--flt=`, `-a`},
	Want:         fftest.Vars{WantParseErrorIs: strconv.ErrSyntax},
},
```

flag_set.go
```diff
@@ -264,19 +256,23 @@ func (fs *FlagSet) parseLongFlag(arg string, args []string) ([]string, error) {
...
 					args = args[1:]
 				}
 			}
 		case !f.isBoolFlag && len(args) > 0:
 			value, args = args[0], args[1:]
-		case !f.isBoolFlag && len(args) <= 0:
+		case !f.isBoolFlag && len(args) == 0:
 			return nil, fmt.Errorf("missing value")
 		default:
 			panic("unreachable")
```
ccoVeille, line 278: The complexity of this code is outstanding

```go
	if value == "" && !eqFound {
		switch {
		case f.isBoolFlag:
			value = "true" // `--foo` defaults to true
			if len(args) > 0 {
				if _, err := strconv.ParseBool(args[0]); err == nil {
					value = args[0] // `--foo false` should also work
					args = args[1:]
				}
			}
		case !f.isBoolFlag && len(args) > 0:
			value, args = args[0], args[1:]
		case !f.isBoolFlag && len(args) == 0:
			return nil, fmt.Errorf("missing value")
		default:
			panic("unreachable")
		}
	}
```

Unless I'm wrong, it could be this

```go
	if value == "" && !eqFound {
		switch {
		case f.isBoolFlag:
			value = "true" // `--foo` defaults to true
			if len(args) > 0 {
				if _, err := strconv.ParseBool(args[0]); err == nil {
					value = args[0] // `--foo false` should also work
					args = args[1:]
				}
			}
		case len(args) > 0:
			value, args = args[0], args[1:]
		default:
			return nil, fmt.Errorf("missing value")
		}
	}
```

flag_set.go
```diff
@@ -264,19 +256,23 @@ func (fs *FlagSet) parseLongFlag(arg string, args []string) ([]string, error) {
...
 		case f.isBoolFlag:
-			value = "true" // `-b` or `--foo` default to true
+			value = "true" // `--foo` defaults to true
 			if len(args) > 0 {
 				if _, err := strconv.ParseBool(args[0]); err == nil {
-					value = args[0] // `-b true` or `--foo false` should also work
+					value = args[0] // `--foo false` should also work
 					args = args[1:]
 				}
 			}
```
ccoVeille, line 272: There is also something strange with the code (the code that already exists)

Let's assume --foo expects a string

--foo leads to an error
--foo true leads to "true"
--foo=1 leads to "1"
--foo=false leads to "false"
--foo= leads to ""

But if --foo is a boolean

--foo false
--foo whatever
--foo 0
--foo 2
--foo=2

Leads to this

--foo=false
--foo whatever
--foo=false
--foo=true 2
an error

This behavior is strange to me, but I'm unsure how other libraries parsing flags do

flag_set.go
telemachus: You are not wrong. Thanks.

flag_set.go
telemachus: The difference between `--foo 2` and `--foo=2` is...not great. That is probably why `flag` in Go's standard library [restricts the form `--flag arg` to non-boolean flags](https://cs.opensource.google/go/go/+/refs/tags/go1.23.4:src/flag/flag.go;l=54).

ccoVeille: Exactly

telemachus: I would be in favor of that here as well, but I think @peterbourgon wants to keep boolean parsing as is. (Also, that would probably be a very breaking API change now. Don't know how that plays with v4 being in beta.)

telemachus: @ccoVeille As a side note, `flag` in the standard library has a parallel inconsistency but for string flags.

```go
	fs := flag.NewFlagSet("whatever", flag.ContinueOnError)
	cfg := struct {
		config  string
		// ...
	}{}
	fs.StringVar(&cfg.config, "config", "", "Use this configuration file")
	// ...
```

```shell
whatever -config=foo   # No error; config is set to "foo"
whatever -config foo   # No error; config is set to "foo"
whatever -config ""    # No error; config is set to ""
whatever -config=""    # No error; config is set to ""
whatever -config       # Error, namely "flag needs an argument: -config"
whatever -config=      # No error; config is set to ""
```

I think that the last two should return the same error, but they do not. I'm guessing this is because the parser would need to do extra work (not much but some) to detect the difference between `-config=` and `-config=""`. I doubt anything can be done about it now (breaking API change), but I think it was a mistake.

ccoVeille: For me, the behavior of the stdlib is fine here. I might look a bit odd, but I'm fine with it.

It has its logic.

About ff lib, I don't know really. You are fixing a bug with the boolean flag after all, so fixing the bug is somehow already breaking something that was broken :smile:, by fixing it.

While your fix is fine, I think the issue of removing the random behavior of boolean flag with a parameter should be considered, at least to have a library that behave like other libraries.

I don't think it was intended, I would remove it. But, except that your PR is fine, you are fixing the behavior with `--foo= ` so nothing about `--foo 0`

So for me it can be merged as is

ccoVeille (approved):

tgulacsi (approved):

telemachus:
@tgulacsi I appreciate the support, but I think we just have to wait for the maintainer to get a chance to review the changes and decide what he wants to do. I don't think that outside people approving will make much a difference one way or the other.

flag_set_test.go
```diff
@@ -83,6 +83,7 @@ func TestFlagSet_Bool(t *testing.T) {
 		{args: []string{"--help"}, wantX: false, wantY: true, wantErr: ff.ErrHelp},
 		{args: []string{"--xflag", "-h"}, wantX: true, wantY: true, wantErr: ff.ErrHelp},
 		{args: []string{"-y", "--help"}, wantX: false, wantY: false, wantErr: ff.ErrHelp},
+		{args: []string{"--xflag=", "--help"}, wantX: true, wantY: false, wantErr: ff.ErrHelp},
```
peterbourgon, line 86: Shouldn't this be

```suggestion
		{args: []string{"--xflag=", "--help"}, wantX: true, wantY: true, wantErr: ff.ErrHelp},
```

?

telemachus: @peterbourgon I see what you mean: `yflag := fs.BoolDefault('y', "yflag", true, "another boolean flag")`. I didn't notice that `yflag` defaulted to true. Sorry.

That said, here's something weird: the tests pass without error in all the cases below.

+ `{args: []string{"--xflag=", "--help"}, wantX: true, wantY: true, wantErr: ff.ErrHelp}`
+ `{args: []string{"--xflag=", "--help"}, wantX: true, wantY: false, wantErr: ff.ErrHelp}`
+ `{args: []string{"--xflag=", "--help"}, wantX: true, wantErr: ff.ErrHelp}`

They all pass. I'm very confused. (I updated to `wantY: true`, but for some reason it doesn't seem to matter.)

peterbourgon: If wantErr is non-nil then the wantX and wantY fields aren't evaluated. My original comment was a bit pedantic, in that sense.

telemachus: Ah, okay. Either way, I clearly need to understand the test structure better.

flag_set.go
ccoVeille: For record, I randomly found a discussion about the exact same issue with boolean flag

- https://github.com/spf13/pflag/issues/410

flag_set.go
peterbourgon: I wrote the code in that form deliberately. I generally like to have `switch` cases follow a kind of [Karnaugh Map](https://en.wikipedia.org/wiki/Karnaugh_map) walk of the state-space, so that each `case` clause can be evaluated independently. Maybe a bit much, but I find it easier to read, and it's no less efficient or whatever than alternatives.

telemachus: @peterbourgon Would you like me to go back to the earlier form? (You can obviously do it yourself too, but let me know.)

cbandy:
While reading the conversations around this, I'm unsure if everyone understands what the shell does with quotes before the Go program receives anything.

For example, this is one argument (word) on the command line that *contains no quotes*: `'asdf'qqq"xyz"''""''gg`

This is a small script that prints each of its arguments on a separate line. With it in `args.sh` plus `chmod +x args.sh`, the above looks like this:
```sh 
#!/bin/sh
printf "|%s|\n" "$@"
```

```console
$ ./args.sh 'asdf'qqq"xyz"''""''gg
|asdfqqqxyzgg|
```

There are no quotes for Go to parse in the following, either. Notice that:
-  `-config ""` arrives as two arguments, 7 bytes then 0 bytes.
- `-config=""` arrives as one argument, 8 bytes ending with `=`.

```console
$ ./args.sh -config=foo -config foo -config "" -config="" -config -config=
|-config=foo|
|-config|
|foo|
|-config|
||
|-config=|
|-config|
|-config=|
```

telemachus:
@cbandy I think I understand your point, but to be clear, do you think that affects the main fix here? That is, even given what you show, I think it's a bug for ff to attach a subsequent argument to `--long=`. To put this another way, regardless of how the shell handles empty strings and quotes, I don't think that ff should treat `[]string["--long=", "foo"]` as if it were `[]string["--long=foo"]`. Do you think I'm wrong?

flag_set.go
```diff
@@ -264,22 +256,24 @@ func (fs *FlagSet) parseLongFlag(arg string, args []string) ([]string, error) {
 		}
 	}
 
-	if value == "" {
+	if eqFound && f.isBoolFlag && value == "" {
+		value = "true" // `--debug=` amounts to `--debug=true`
+	}
```
cbandy, line 261: I would move this into the switch, if possible.

```go
	switch {
	case hasEquals && f.isBoolFlag && value == "":
		// Treat --bool= the same as --bool=true
		value = "true"

	case hasEquals:
		// Use the value within this arg as-is.

	case f.isBoolFlag:
		// Use the value of the next arg, if it exists and looks related.
		...

	default:
		// Use the value of the next arg, if it exists.
		...
	}

cbandy:
> I don't think that ff should treat `[]string["--long=", "foo"]` as if it were `[]string["--long=foo"]`.

Agreed.

cbandy (commented):
Tests LGTM.

flag_set.go
telemachus: I'm happy to look at that, but I'm not going to push any further changes until we hear from @peterbourgon. I'd like to know what he's thinking about this PR before changing it further.

PR #141, opened 2025-01-28
Allow long flag with equal sign but no args (--foo= --bar=x)

### LINKED ISSUE #138: Questions about parsing of long flags with = in them
(First, thanks for ff.) I have some questions about how ff.FlagSets parse long flags that use `=`. After looking at the code, I'm not sure whether I'm seeing a bug or just a feature that I didn't expect.

1. Boolean flags that are passed as `--flagname=` (that is, with a long flag and no value after `=`) are parsed as true. Based on [this comment](https://github.com/peterbourgon/ff/blob/main/flag_set.go#L270), you may be doing this deliberately, but I'm inclined to think that this is a bug resulting from how you handle `=` earlier in `parseLongFlag`.
2. Similarly, all other flag types that omit a value after `=` consume the next `arg` as their value. This can have very surprising results for users who mistakenly forget a value after `=`. Consider the following.
```go
	fs := ff.NewFlagSet("myprogram")
	debug := fs.Bool('d', "debug", "log debug information")
	file := fs.String('f', "file", "", "file to frobnicate")

	if err := fs.Parse([]string{"--file=", "--debug="}); err != nil {
		panic(err)
	}

	fmt.Printf("file=%q\n", *file)
        fmt.Printf("debug=%v\n", *debug)
	fmt.Printf("args = %+v", fs.GetArgs())
        // Output:
        // file="--debug="
        // debug=false
        // args = []
```

By way of comparison, stdlib's `flag` treats all empty values after `=` as errors except for string flags. (Presumably because they only return an error if a `strconv` conversion fails, and there is no conversion in the case of string values. An empty value == `""`.) This strikes me as the right approach (maybe even string flags should fail here). Leaving out a value is presumably user error, and parsing should fail and stop. (Alternatively, the treatment of empty values with `=` should be documented.)

If this is a bug, I think I have a fix, and I'm happy to make a PR. Let me know what you think.

tgulacsi:
Allow empty long flags (--tls.key= --tls.cert=) - the current parser uses the next argument (--tls.cert=) as the value.

Fixes #138 .

### DIFF
--- a/flag_set.go
+++ b/flag_set.go
@@ -242,12 +242,14 @@ func (fs *FlagSet) parseShortFlag(arg string, args []string) ([]string, error) {
 
 func (fs *FlagSet) parseLongFlag(arg string, args []string) ([]string, error) {
 	var (
-		name  string
-		value string
+		name         string
+		value        string
+		hasEqualSign bool
 	)
 
 	if equals := strings.IndexRune(arg, '='); equals > 0 {
 		arg, value = arg[:equals], arg[equals+1:]
+		hasEqualSign = true
 	}
 
 	name = strings.TrimPrefix(arg, "--")
@@ -264,7 +266,7 @@ func (fs *FlagSet) parseLongFlag(arg string, args []string) ([]string, error) {
 		}
 	}
 
-	if value == "" {
+	if value == "" && (f.isBoolFlag || !hasEqualSign) {
 		switch {
 		case f.isBoolFlag:
 			value = "true" // `-b` or `--foo` default to true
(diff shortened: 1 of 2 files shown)


tgulacsi:
Thanks @ccoVeille , you're right, a86c7b7195e25ab598979445fcc658e82e67c48e fixes my problem (long string args with default values where I want to specify an empty string).
When will it be merged & released ?

ccoVeille:
I don't know.

I'm a random Gopher just like you.

The repository owner didn't reply to the question we raised in [this thread](https://github.com/peterbourgon/ff/pull/139#discussion_r1920717627)

tgulacsi:
Let's concentrate on #139

PR #143, opened 2025-03-15
Fix custom types in structs

peterbourgon:
ref: #136


peterbourgon:
Done properly in #148 I think

PR #156, opened 2025-08-28
Reduce dependency graph by bumping dependencies

### LINKED ISSUE #154: Dependency gopkg.in/yaml.v2 is now unmaintained
I just noticed that ff depends on gopkg.in/yaml.v2 which has been archived on April 1st. So please consider switching to another fork.

It looks like one possible candidate is https://github.com/yaml/go-yaml which looks promising but it's really up to you.

cbandy:
Updating these two direct dependencies reduces the number of indirect dependencies to almost zero. The `gopkg.in/yaml.v2` module has moved to `go.yaml.in/yaml/v2` and is now maintained by the YAML organization on GitHub.

Fixes: #154
See: https://www.github.com/pelletier/go-toml/issues/872
See: https://www.github.com/yaml/go-yaml/discussions/11


cbandy:
```console
$ (main) go mod graph | wc -l
23
$ (fewer-deps) go mod graph | wc -l
6

peterbourgon:
I'm happy to dig into this but just to be clear "the number of indirect dependencies" is not a number that needs to be optimized for 😇 The ff module and its deps is always imported by something else, those importers will specify transitive dep versions as they require, and it is those end-user go.mods which dictate the ultimate compilation graph, not anything in ff itself.

cbandy:
True! I find things like this help with unsophisticated scanners in my day job: licenses, vulnerabilities, and such.

Certainly no urgency from me on this. 👍🏻

